PR #4, opened 2025-03-22
add authentication interface for login

Cazden:
Adds initial authentication layout, auth UI components and routing setup for Login, Password and 2FA pages.


src/pages/auth/Login.tsx
```diff
@@ -1,7 +1,108 @@
...
+
+  function onSubmitEmail() {
+    void navigate({ to: "/auth/password" })
+  }
+
+  function onSubmitSSO() {
+    // TODO: Handle SSO login
+  }
+
+  const renderEmailForm = () => (
```
ezekg: Is there any particular reason this is a function and not a variable?

src/pages/auth/Password.tsx
```diff
@@ -0,0 +1,137 @@
...
+      remember: false,
+    },
+  })
+
+  function onSubmitPassword() {
+    // TODO: Handle auth
+    void navigate({ to: "/auth/verify" })
+  }
+
+  const renderPasswordForm = () => (
```
ezekg: Same here. Not sure what the purpose of the function is, but lmk if it's intentional.

src/pages/auth/Verify.tsx
```diff
@@ -0,0 +1,252 @@
...
+                    name={fieldName}
+                    render={({ field }) => (
+                      <FormItem>
+                        <FormControl>
+                          <Input
+                            ref={(el) => {
+                              inputRefs.current[i] = el
+                            }}
+                            aria-invalid={error ? "true" : "false"}
+                            type="text"
```
ezekg: I think we can also add `pattern="\d"` so that mobile devices only show the numeric keyboard.

ezekg (commented):
Few comments but overall looks great. I do like that we dropped semicolons lol.

src/pages/auth/Login.tsx
Cazden: Originally there were two functions in the Login component—renderEmailForm and renderPasswordForm—and I was conditionally rendering these separate functions for easier readability in the final JSX based on which step the user was on. I ended up separating them into different routes instead due to an internal react component id issue, so this was basically a leftover artifact from that refactor and got overlooked, so I'll fix this.

src/pages/auth/Password.tsx
Cazden: See previous comment, it's the same situation.

src/pages/auth/Verify.tsx
Cazden: Sure sounds good.

src/pages/auth/Verify.tsx
```diff
@@ -0,0 +1,252 @@
...
+  digit5: z.string().regex(/^\d$/, "Must be one digit"),
+  digit6: z.string().regex(/^\d$/, "Must be one digit"),
+})
+
+export default function Verify() {
+  const [error, setError] = useState<string | null>(null)
+
+  const navigate = useNavigate()
+
+  const verificatonForm = useForm<z.infer<typeof verificationSchema>>({
```
ezekg, line 33: Is there a way to auto-focus the first 2FA field so a user doesn't have to click the input before typing?

ezekg: Also doing the same for the password field may be helpful too.

Cazden:
Refactored functions in Login and Password components to inline JSX. Also added mobile numeric pattern, OTP auto complete and auto focus for password and OTP fields.

ezekg (approved):
Autofocus works great. Re: mobile design, lmk when it's added and I'll rereview.

ezekg:
Let's fix the alignment of the back button so that it's inline with the logo again:

image

And then let's fix the layout on landscape mobile so that it doesn't overlap with the logo and scrolls up/down instead:

image

I think the overlap happens because the logo is absolute positioned. Might help to change that so that it's relative and centered, and build the header in a way that other items, like the back button or nav, and other items like the login form, all flow nicely on the page when they resize and don't overlap with each other.

In general, staying away from absolutely positioning large components (but e.g. tooltips, etc. are fine) will help things be responsive with less work. You can align and center things using flex/grid instead of using absolute positioning.

Cazden:
I removed the absolutely positioned nav elements so they all flow together and there's no overlap. Window is also scrollable in landscape for screens that can't fit the content height, and the back button has been repositioned back to the top-left for desktop, and above the form on mobile to be closer to the thumb zone.

Mobile:
Mobile

Mobile Landscape:
image

Desktop:
image

src/pages/auth/Login.tsx
```diff
@@ -1,7 +1,101 @@
...
+              </FormItem>
+            )}
+          />
+          <Button type="submit" size="lg" className="w-full">
+            Continue
+          </Button>
+        </form>
+      </Form>
+
+      <div className="space-x-2 rounded border border-content-subdued p-3 text-center text-sm select-none">
```
ezekg: Nit: let's use p-2 here so that the call to action's size matches the submit button's.

```suggestion
      <div className="space-x-2 rounded border border-content-subdued p-2 text-center text-sm select-none">
```

ezekg (approved):
One nit. LGTM.

PR #5, opened 2025-03-28
add basic interface for registration

Cazden:
Adds a basic account registration form interface, which includes an email and password field, an auto-generated slug and a placeholder API call on submit.

### DIFF
--- a/src/components/Loading/Dots.tsx
+++ b/src/components/Loading/Dots.tsx
@@ -1,0 +1,49 @@
+import { useState, useEffect } from "react"
+
+interface DotsProps {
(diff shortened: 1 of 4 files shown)


Cazden:
I added loading dots to the button while the form is submitting. By default, the dots component will be brand green, but for the button here I matched the dots to the background color.




And here's some form previews:
Mobile

Mobile

Cazden:
Added a new Company ID field that will display when a company ID already exists for the entered domain. Wired up with a dummy function for now.

image

ezekg (approved):
LGTM

PR #7, opened 2025-03-31
add single sign-on interface

Cazden:
Adds initial single-sign on interface and routing, ready for implementing SSO logic.

### DIFF
--- a/src/routes/auth.external.tsx
+++ b/src/routes/auth.external.tsx
@@ -1,0 +1,6 @@
+import { createFileRoute } from "@tanstack/react-router"
+import * as Page from "@pages/index"
+
+export const Route = createFileRoute("/auth/external")({
+  component: () => <Page.Auth.External />,
+})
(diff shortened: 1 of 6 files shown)


src/routes/auth.external.tsx
```diff
@@ -0,0 +1,6 @@
+import { createFileRoute } from "@tanstack/react-router"
+import * as Page from "@pages/index"
+
+export const Route = createFileRoute("/auth/external")({
```
ezekg: Let's keep this route named `/auth/sso` for clarity (using `/external` isn't very clear imo). That way we can eventually add e.g. social login, which is different than SSO, but still external.

ezekg (approved):
This looks good. One comment on naming to clarify purpose/intent. And just FYI — this can/will mostly be automated since the API currently supports this flow:

1. User enters email (i.e. step 1 of the 'general' login flow we've already covered).
2. API returns either `PASSWORD_REQUIRED` or `SSO_REQUIRED`, indicating next steps.
3. UI redirects to either the password flow or a provided SSO URL.
5. ???
6. Profit (i.e. successful login).

We can still have this page ofc since some non-techies will like the explicit flow.

src/routes/auth.external.tsx
Cazden: Sounds good, I've updated it from external to sso.

PR #8, opened 2025-04-23
add authentication api integration

Cazden:
Adds authentication with Keygen API for auth routes & components, environment setup/config for accessing a locally hosted API, and routing/path setup for a tenant-specific portal (WIP).


.gitignore
```diff
@@ -23,4 +23,6 @@ dist-ssr
 *.sln
 *.sw?
 
-**/routeTree.gen.ts
\ No newline at end of file
+**/routeTree.gen.ts
+
+.env
```
ezekg: Good practice to do this so that env files are never accidentally leaked while allowing a sample file:

```
.env*
!.env.sample
```

src/keygen/authenticate.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (!response.ok) {
+      const result = (await response.json()) as AuthResponse
+      const codes = result.errors?.map((err) => err.code) || []
+
+      if (
+        codes.includes("PASSWORD_REQUIRED") ||
+        codes.includes("OTP_REQUIRED") ||
+        codes.includes("EMAIL_INVALID") ||
+        codes.includes("PASSWORD_INVALID") ||
+        codes.includes("OTP_INVALID")
```
ezekg: I feel like this could be cleaned up using a generalized enum of API error codes, and a set of auth error codes:

```ts
enum ErrorCode {
  PASSWORD_REQUIRED = "PASSWORD_REQUIRED",
  PASSWORD_INVALID = "PASSWORD_INVALID",
  EMAIL_INVALID = "EMAIL_INVALID",
  OTP_REQUIRED = "OTP_REQUIRED",
  OTP_INVALID = "OTP_INVALID",
}

const authErrorCodes = new Set<ErrorCode>([
  ErrorCode.PASSWORD_REQUIRED,
  ErrorCode.PASSWORD_INVALID,
  ErrorCode.EMAIL_INVALID,
  ErrorCode.OTP_REQUIRED,
  ErrorCode.OTP_INVALID,
])

const codes = result.errors?.map(err => err.code) || []
if (codes.some(code => authErrorCodes.has(code as ErrorCode))) {
  // ...
}
```

src/keygen/authenticate.ts
```diff
@@ -0,0 +1,96 @@
...
+        Authorization: `Basic ${credentials}`,
+        "Content-Type": "application/vnd.api+json",
+        Accept: "application/vnd.api+json",
+        "Keygen-Version": config.version,
+      },
+      body,
+    })
+
+    if (!response.ok) {
+      const result = (await response.json()) as AuthResponse
```
ezekg: You could move this up to remove the duplication after this conditional.

src/keygen/verify.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Checks user token validity with the Keygen API.
+ * If the response returns a 401, the token is invalid, otherwise it's still valid.
+ *
+ * @param {string} props.token - Token to verify
+ * @return {Promise<boolean>}
+ */
+export async function verify({ tokenId, token }: VerifyProps): Promise<boolean> {
+  try {
+    const response = await fetch(`https://${config.host}/v1/accounts/${config.id}/tokens/${tokenId}`, {
```
ezekg: Eventually, we may want to use [`/v1/accounts/<account>/me`](https://keygen.sh/docs/api/profiles/) to save a request when obtaining the current user. We can leave this for now and handle this later — just making a note here for the future.

tsconfig.json
```diff
@@ -10,7 +10,12 @@
       "@/*": ["src/*"],
       "@assets/*": ["src/assets/*"],
       "@components/*": ["src/components/*"],
+      "@contexts/*": ["src/contexts/*"],
+      "@hooks/*": ["src/hooks/*"],
+      "@keygen/*": ["src/keygen/*"],
       "@layouts/*": ["src/layouts/*"],
+      "@providers/*": ["src/providers/*"],
+      "@guards/*": ["src/guards/*"],
```
ezekg: I'm not sure I like these `@foo` aliases because they could end up conflicting with npm packages in the future, since npm uses `@` for scoped packages. I think I'd rather use a global `~/` or `@/` alias for `src/*` and leave it at that (not personally a fan of the `@/` convention, which grew out of Webpack quirks, and I tend to prefer `~/` but won't die on that hill).

ezekg (changes requested):
Overall looks good. Have some minor notes.

src/keygen/authenticate.ts
ezekg: In general, string constants like this should probably be pulled up into a const or an enum.

tsconfig.json
```diff
@@ -8,11 +8,7 @@
     "baseUrl": ".",
     "paths": {
       "@/*": ["src/*"],
-      "@assets/*": ["src/assets/*"],
-      "@components/*": ["src/components/*"],
-      "@layouts/*": ["src/layouts/*"],
-      "@pages/*": ["src/pages/*"],
-      "@routes/*": ["src/routes/*"]
+      "@assets/*": ["src/assets/*"], // shadcn
```
ezekg: Hate to be nitpicky but let's go ahead and nix this too and move everything from `src/assets` into `src`. The `assets` name isn't descriptive, because there are components in there not just e.g. image assets, and naming it e.g. `vendor` doesn't add much either since everybody knows `components/ui` is vendored shadcn. We'll eventually need our own `lib` anyways, so it just makes sense to flatten these imo and follow what others do. Fine with keeping logos under assets if you want (arguably they should probably be under `/public` since they're static assets and you can just link to them directly e.g. `src='/logo.svg'`).

ezekg: e.g.: https://github.com/Kiranism/next-shadcn-dashboard-starter

Cazden: Perfect. I went ahead and renamed all file names to kebab-case since we're moving shadcn's components into our own component folder so that everything is consistent across the codebase, since shadcn also uses kebab-case.

ezekg (approved):

PR #10, opened 2025-05-12
add app sidebar interface

Cazden:
Adds base app sidebar interface, SVGR for importing SVG icons as React components for styling cases, i.e. hover states for sidebar icons, and sets up future Sidebar additions such as CmdK, searching and app navigation.


Cazden:
image
image
image

Cazden:
Sidebar UI overview:
image

Rail hover:
image

Command menu:
image

Search menu:
image

Product/environment menu:
image

Cazden:
Portrait:


Landscape:

src/components/sidebar/combobox.tsx
```diff
@@ -0,0 +1,130 @@
...
+      { id: "prd_002_stage", label: "Staging" },
+      { id: "prd_002_prod", label: "Production" },
+    ],
+  },
+]
+
+/**
+ * Renders a combobox that allows users to select a product and environment.
+ *
+ * @returns {JSX.Element}
```
ezekg: I think instead of using docblocks, you should type the function's return value so that we get the type safety via the typescript compiler. We don't really need to mix jsdoc docblocks and typescript — just feels kind of redundant tbh. I say keep the comment when it's helpful but other than that, don't go out of your way to use docblocks because they're just going to end up out of date as the project evolves. Instead, opt for actual typing instead of inference.

Or if you do want to use docblocks, consider [tsdoc](https://tsdoc.org/) which removes type duplication.

src/keygen/logout.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Logout function that removes token from storage and redirects.
+ * (Cazden) To be expanded upon later.
```
ezekg: Conventional way for these self-comments are `FIXME(cazden)`, `TODO(cazden)`, etc., just FYI.

Easier to search if you're ever bored.

ezekg (approved):

ezekg:
Looks good! Couple comments but no changes requested. Feel free to merge. 🤘

src/components/sidebar/combobox.tsx
Cazden: Awesome good to know, makes total sense. I pushed a change that cleans up the jsdoc and adds return types.

PR #12, opened 2025-06-02
add environments api integration

Cazden:
Adds integration with Keygen's environment CRUD API endpoints, and removes placeholder environments data. Also includes some housekeeping such as server and auth types being moved into a new types folder and refactored accordingly, and some minor UI changes to the environment modals to accommodate real data.


src/constants/auth.ts
```diff
@@ -0,0 +1,9 @@
+enum AUTH_ERROR_CODES {
```
ezekg: Nit: enum names should be UpperCamelCase, and properties should be CONSTANT_CASE (like they are currently). At least that's what's typical in TS code bases I've worked on. Making them constant-case conveys (to me) that they're a constant scalar value like a string or boolean, like an env var or something.

src/keygen/environments/create.ts
```diff
@@ -0,0 +1,72 @@
+import config from "@/keygen/config"
+import { EnvironmentResponse } from "@/types/environments"
+import { STRATEGIES } from "@/constants/environments"
```
ezekg: These constants aren't defined, i.e. the file is missing in the PR. Personally, based on [my previous comment](https://github.com/keygen-sh/keygen-portal/pull/12/files#r2124875351), I'd throw these into `/types` like where `EnvironmentResponse` is and nix the constants subdir unless there's actually a large amount of real constants to define.

src/keygen/environments/get.ts
```diff
@@ -0,0 +1,54 @@
...
+}: GetProps): Promise<EnvironmentResponse> {
+  try {
+    const response = await fetch(
+      `https://${config.host}/v1/accounts/${config.id}/environments/${id}`,
+      {
+        method: "GET",
+        headers: {
+          Authorization: `Bearer ${token}`,
+          Accept: "application/vnd.api+json",
+          "Keygen-Version": config.version,
```
ezekg: Eventually might be a good idea to abstract these API calls out so that you don't have to repeat yourself i.r.t. required headers. And it'll make adding new headers easier later on, e.g. `Keygen-Environment`. Like abstract out into a `Client` and create a singleton instance or smth that uses the config, auth context e.g. current account/token/user, etc. Then use methods like `client.tokens.create`, etc. in these resource files.

This also allows you consolidate error handling, e.g. your `CLIENT_ERROR` bit.

src/keygen/environments/remove.ts
```diff
@@ -0,0 +1,59 @@
...
+
+interface RemoveProps {
+  token: string
+  id: string
+}
+
+export default async function remove({
+  token,
+  id,
+}: RemoveProps): Promise<EnvironmentResponse> {
```
ezekg: I'd make this clear that it returns `null` on success, since the delete endpoint returns a `204 No Content`.

src/keygen/environments/update.ts
```diff
@@ -0,0 +1,70 @@
...
+  id,
+  name,
+  code,
+}: UpdateProps): Promise<EnvironmentResponse> {
+  try {
+    const body = {
+      data: {
+        type: "environments",
+        attributes: {
+          ...(name && { name }),
```
ezekg: Never seen this before. Nice trick. 😆

src/keygen/environments/create.ts
Cazden: Interesting, it's there now since I moved them into the types file.

Cazden:
I added the changes you mentioned and I think we're ready to merge so I'm gonna switch out of draft. I think modal transitions need a bit of work, and I need to make sure everything is responsive for mobile, but that's UI-related so I'll start another branch for that after this one is merged.

src/types/environments.ts
```diff
@@ -1,11 +1,56 @@
...
+  LIST = "list",
+  DETAILS = "details",
+}
+
+export enum IsolationStrategies {
+  ISOLATED = "ISOLATED",
+  SHARED = "SHARED",
+}
+
+export enum EnvironmentDescriptions {
```
ezekg: Nit: it's conventional for enums to be singular, not plural.

src/hooks/use-session.tsx
```diff
@@ -0,0 +1,61 @@
...
+        keygen.client.setToken(token)
+
+        navigate({
+          to: "/$id/app/dashboard",
+          params: { id: keygen.config.id },
+        })
+
+        return
+      } catch (error) {
+        console.error(error)
```
ezekg, line 35: Note to self: we should eventually integrate Sentry for exception tracking.

ezekg (approved):
Looking good. One nit, but it isn't a blocker and can be done in the next PR.

LGTM.

PR #13, opened 2025-06-10
add responsiveness to environment modal

Cazden:
Adds a responsive environments interface for both landscape/portrait modes on mobile screens, and some other minor UI changes/fixes.


src/components/environments/edit/edit-form.tsx
```diff
@@ -71,106 +84,166 @@ export default function EnvironmentEditForm({
...
+                  )}
+                />
+              </CardContent>
+            </Card>
           </div>
-        </div>
+        </ScrollArea>
+
+        <p className="hidden flex-wrap items-center p-4 text-sm text-content-subdued md:flex">
+          To learn more about environments, see the&nbsp;
```
ezekg: Instead of `&nbsp;`, you can do `{' '}` to force a space in React.

src/components/environments/view/details.tsx
```diff
@@ -115,15 +137,6 @@ export default function EnvironmentDetails({
               />
             </div>
           </CollapsibleMenu>
-          <CollapsibleMenu title="Tokens">
-            <div className="mb-3 flex items-center justify-between">
-              <h2 className="text-sm font-semibold">Environment Tokens</h2>
```
ezekg, line 120: Are we going to add this later, or put it all under a Tokens page instead?

ezekg (approved):
Couple notes, but looks good. LGTM.

src/components/environments/view/details.tsx
Cazden: When I removed this I was thinking a tokens page to keep everything central, but it might be helpful to have it in both places so users don't have to switch context, or get confused if they expected tokens to be handled within the modal. What do you think?

ezekg: I think it'll be fine to have a dedicated page for it.

ezekg: Can always use the command bar to generate a new token quickly after creating a new environment.

PR #15, opened 2025-07-23
add products api integration

Cazden:
Adds integration with Keygen's product CRUD API endpoints and removes placeholder products data.


ezekg (changes requested):
Love it. A couple notes:

1. When I created a new product with a `CLOSED` distribution strategy, the resulting product's strategy was the default, `LICENSED`. Edit works fine. Maybe an issue with the 2-step form not keeping state?
2. After create, can we redirect to the new product instead of staying at the list view?

src/components/sidebar/combobox.tsx
```diff
@@ -19,50 +21,81 @@ import {
 
 import { Droplet } from "lucide-react"
 
+import { Product } from "@/types/products"
```
ezekg: Environments are global and don't need to be grouped by product. Just list out the environments, with no hierarchy.

src/hooks/use-environment.tsx
```diff
@@ -0,0 +1,82 @@
...
+    queryFn: () =>
+      keygen.environments.list({}).then((response) => response.data ?? []),
+  })
+}
+
+export function useCreateEnvironment() {
+  const queryClient = useQueryClient()
+
+  return useMutation({
+    mutationFn: (payload: any) =>
```
ezekg: Would love to see a type here instead of `any` (same for other mutation functions)

ezekg:
There's an old file sitting around that doesn't seem to be used anymore: https://github.com/keygen-sh/keygen-portal/tree/master/src/components/Loading

src/components/sidebar/combobox.tsx
```diff
@@ -12,58 +10,48 @@ import {
...
-  const productOptions = useMemo(() => {
-    return products.map((product: Product) => ({
-      id: product.id,
-      name: product.attributes.name,
-    }))
-  }, [products])
+  const [openModal, setOpenModal] = useState(false)
+  const [openPopover, setOpenPopover] = useState(false)
+  const { data: environments = [], isLoading: environmentsLoading } =
+    useQueryEnvironments()
```
ezekg: Nit: this is typically destructured like below, but maybe Prettier is different now and if so, disregard.

```js
const {
  isLoading: environmentsLoading,
  data: environments = [],
} = useQueryEnvironments()
```

src/hooks/use-query-environment.ts
```diff
@@ -3,7 +3,7 @@ import { Environment } from "@/types/environments"
 import * as keygen from "@/keygen"
 import { diff } from "@/lib/utils"
 
-export function useEnvironment(environmentId: string) {
+export function useQueryEnvironment(environmentId: string) {
```
ezekg: I think we should put queries under a subdirectory to keep them separate from other hooks, instead of naming these like `use-query-xxx`, which is a bit ambiguous imo because it reads like we're using a "query environment" which doesn't grok the actual use of "querying an environment." Maybe `use-xxx-query` would be clearer, but idk. This thought mainly stems from the conflict between the environment hooks.

tbh I'd name the hooks like `useReadEnvironment` and `useReadEnvironments` since that way the hooks follow a CRUD naming convention (i.e. create, read, update, delete). If we put them under `src/hooks/queries`, it'd then be clearer they're queries from the import path alone now.

Here is prior work on that front showing it's a pattern:

- https://github.com/polarsource/polar/tree/main/clients/apps/web/src/hooks/queries
- https://github.com/shapeshift/web/tree/main/src/hooks/queries

Or use `src/queries` to make it even clearer? I kind of like the idea of separating data hooks from normal hooks. There's prior work on that front as well:

- https://github.com/Crossbell-Box/xLog/tree/main/src/queries
- https://github.com/windup/windup/tree/master/pf-ui/src/main/webapp/src/queries

lmk what you think?

ezekg (changes requested):
Looks good. A new notes on maybe separating hooks and queries.

Also, I pushed a fix for the casing issue.

src/hooks/use-query-environment.ts
Cazden: Makes total sense. I like the idea of separating the queries into their own folder separate from regular hooks. To keep import paths slim, I'll still group them in a single file, i.e. `useReadEnvironment`, `useCreateEnvironment`, etc. -> `src/queries/environments.ts`

src/components/sidebar/combobox.tsx
Cazden: Yeah this seems to be a Prettier thing.

src/queries/products.ts
```diff
@@ -51,17 +52,18 @@ export function useCreateProduct() {
         .then((response) => response.data as Product),
 
     onSuccess: (newProduct) => {
-      queryClient.setQueryData<Product[]>(["products"], (old) =>
+      queryClient.setQueryData<Product[]>(["products", code], (old) =>
```
ezekg: Let's do query keys like so we're free to add more filters, etc. in the future.

```js
["products", { environment, page, /* etc. */ }] // collection
["product", productId, { environment }] // individual
```

Ref: https://tanstack.com/query/v5/docs/framework/react/guides/query-keys

src/types/relationships.ts
```diff
@@ -0,0 +1,29 @@
...
+  self?: string | null
+}
+
+export type Relationship<T = Resource | Resource[]> = {
+  data?: T | null
+  links?: Links
+  meta?: Record<string, unknown>
+} | null
+
+export type Relationships = {
```
ezekg: This catch-all relationship bit is going to bite us, so I'd rather be resource-specific since every resource has different relationships. They all have an `account` relationship, but other than that, it'll depend on the resource.

src/types/relationships.ts
```diff
@@ -0,0 +1,29 @@
+export interface Resource {
```
ezekg: These shared `data` and `id` props — especially i.r.t. relationships — are called "linkage" in [the JSONAPI spec](https://jsonapi.org/), which we loosely follow, and I use that "linkage" terminology in the API and docs, so let's use that here too.

A "resource" is a full object, e.g. attributes, relationships, etc.

src/providers/environment-provider.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [code, setCode] = useState<string | null>(null)
+  const queryClient = useQueryClient()
+
+  const getEnvironmentToken = useCallback(async (id: string, code: string) => {
+    if (cache[id]) return cache[id]
+
+    keygen.client.setEnvironment(code)
+
+    const response: any = await keygen.client.request(
+      `/accounts/${keygen.config.id}/tokens`,
```
ezekg: Eventually, this should probably be moved to a token query, e.g. `useCreateEnvironmentToken`, which would itself probably be a variant of a `useCreateToken` query hook.

src/queries/products.ts
```diff
@@ -0,0 +1,115 @@
...
+  Product,
+  CreateProductPayload,
+  UpdateProductPayload,
+} from "@/types/products"
+import { ErrorResponse } from "@/types/api"
+
+import * as keygen from "@/keygen"
+import { diff } from "@/lib/utils"
+
+export function useReadProduct(productId: string) {
```
ezekg: Feel free to change this to get/list if that's what you prefer. Saw that naming convention here:

https://github.com/keygen-sh/keygen-portal/blob/305df35c5aba9367b4e6c196cf1dcaa2e41ca26e/src/keygen/products/index.ts#L2-L3

I'm fine either way, but just wanted to note that you're free to name w/e.

ezekg (approved):
A few notes/nits. Feel free to address in the next PR, or ignore.

ezekg (approved):

src/types/tokens.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export type Token = Resource<"tokens", TokenAttributes, TokenRelationships>
+
+export type TokenResponse = APIResponse<Token>
+export type TokensListResponse = APIResponse<Token[]>
+
+export type CreateTokenPayload =
+  | {
+      environment: Exclude<TokenRelationships["environment"], undefined>
```
ezekg, line 38: What does `Exclude<T, undefined>` do?

ezekg (approved):
Looks good. LGTM!

src/types/tokens.ts
Cazden: `environment` and `bearer` are optional in `TokenRelationships` so we can switch between the union types in `CreateTokenPayload` without any complaints, and what this is essentially doing is making whichever one we need required by excluding undefined.

PR #16, opened 2025-08-27
add create policy interface

Cazden:
Adds policies user interface, including pages for policies list and policy details, dialog for advanced policy info, various component additions and refactors, and create/update policy flow with templates for various policy parameters. Some WIP, i.e. the create/update flow is not immediately ready, but will be included once the pull request is ready. Opening this now to share the current list/details pages for review.

image

image

image

image

image

image

image

image


src/lib/policies.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const p = policy.attributes.maxProcesses
+  return typeof p === "number" && p > 0
+}
+
+export function isLeaseBased(policy: Policy): boolean {
+  return policy.attributes.requireHeartbeat === true
+}
+
+export function isFeatureBased(policy: Policy): boolean {
+  const count = policy.relationships?.entitlements?.data?.length ?? 0
```
ezekg: The entitlement relationship will always exist, but it will never contain `data`, because it's a plural relationship (and it could contain any number of records, so they aren't loaded for performance reasons). Entitlements have to be fetched separately by following the relationship link in `data.relationships.entitlements.links.related`.

src/lib/policies.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export function isUsageBased(policy: Policy): boolean {
+  const uses = policy.attributes.maxUses
+  return typeof uses === "number" && uses > 0
+}
+
+export function policyDurationToString(seconds?: number | null): string {
+  if (seconds == null || seconds === 0) return "Unlimited"
+  const match = PolicyDurations.find((duration) => duration.seconds === seconds)
```
ezekg: I would use something like https://date-fns.org/v4.1.0/docs/formatDuration for this.

ezekg: Mainly saying this because I want to allow freeform durations as well because some customers have wanted to set it to some arbitrary number based on their business requirements. So having presets + custom would be ideal.

src/lib/policies.tsx
```diff
@@ -0,0 +1,122 @@
...
+  tooltip?: string
+}
+
+export function formatPolicyAttribute(
+  key: keyof Policy["attributes"],
+  raw: unknown,
+  options?: { forceDisabled?: boolean },
+): Formatted {
+  const isUnset =
+    raw === null || raw === undefined || (typeof raw === "string" && raw === "")
```
ezekg: Tip `raw == null` matches both null and undefined. Useful when checking for both.

This could be simplified to `raw == null || raw === ""`. 🙂

src/lib/policies.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+
+  let value: string
+  if (key === "duration") value = policyDurationToString(raw as number)
+  else if (key === "created" || key === "updated")
+    value = formatDate(String(raw))
+  else if (typeof raw === "boolean") value = raw ? "Enabled" : "Disabled"
+  else if (typeof raw === "string") value = policyLabel(raw)
+  else value = String(raw)
```
ezekg: All of this should probably be extracted out and handled in a separate component that has an explicit `type` prop so that you don't need to determine formatting based on the value alone (nor basing format on key):

```tsx
<Attribute type="duration" label="Duration" value={attrs.duration} tooltip={/* ... */} />
<Attribute type="datetime" label="Created" value={attrs.created} />
<Attribute type="string" label="Name" value={attrs.name} />
<Attribute type="code" label="Code" value={attrs.code} />
<Attribute type="json" label="Metadata" value={attrs.metadata} />
```

I'm sure this kind of logic will need to be used by other resources.

src/lib/policies.tsx
Cazden: Thanks for pointing that out, I'll fix this.

src/lib/policies.tsx
Cazden: Yeah I think that would be a better approach. This whole function has been on my list of things to refactor, along with refactoring the components that are using it. A new component under `src > components > attribute` (i.e. `<Attribute.Value />`) is probably where I'll implement your suggestion.

Cazden:
Here's some previews of the new duration input that allows the user to select a unit type for custom durations, or choose presets from a dropdown.

image

image

And some previews of the create flow. There's still some things that need to be done here (i.e. the tooltip icon placement for text wrapped fields looks bad, some form validation, etc.):

image

image

image

image

image

image

image

image

Cazden:
Added a mobile-friendly step indicator and collapsed badges into a popover:

image

Cazden:
Here's some previews of the scratch flow:

image
image
image

Cazden:
Renamed the pull request for clarity, since this doesn't include the update flow. I'd also like to bring attention to the form validation and form defaults in this PR; let me know if anything looks off to you or needs extending, i.e. in the scratch flow, if a user enables `Require heartbeat`, do we need to validate all subsequent lease-based fields? And do you want to autofill them on enable? etc. just some examples of things to look at, so let me know if you have any specific comments towards things like that.

src/components/duration-input.tsx
```diff
@@ -0,0 +1,234 @@
...
+          <Input
+            ref={inputRef}
+            type="number"
+            min={1}
+            step={1}
+            value={isUnlimited ? "" : num || ""}
+            onFocus={() => setPresetsOpen(true)}
+            onChange={onNumberChange}
+            disabled={disabled || isUnlimited}
+            className="rounded-r-none"
```
ezekg, line 216: Hide the up/down arrows on numeric fields. This is in Firefox.

Image

src/components/duration-input.tsx
```diff
@@ -0,0 +1,234 @@
...
+        </PopoverContent>
+      </Popover>
+
+      <Popover open={unitsOpen} onOpenChange={setUnitsOpen}>
+        <PopoverTrigger asChild>
+          <Button
+            type="button"
+            variant="outline"
+            disabled={disabled}
+            className="min-w-24 rounded-l-none border-l-0 bg-background-1 text-content-muted"
```
ezekg: Rather than equal width, I'd rather see the "day" button take the minimal amount of room, e.g.:

Image

I think you can do this by removing the min width.

src/components/policies/fields/timed.tsx
```diff
@@ -0,0 +1,523 @@
...
+              tooltip={PolicyAttributeDescriptions.renewalStrategy}
+            >
+              <Select
+                value={field.value ?? ""}
+                onValueChange={(value) =>
+                  field.onChange(
+                    value === "" ? undefined : (value as RenewalStrategy),
+                  )
+                }
+                disabled
```
ezekg: This field is always disabled:

Image

src/components/policies/fields/node-locked.tsx
```diff
@@ -0,0 +1,568 @@
...
+
+      <FormField
+        control={form.control}
+        name="machineUniquenessStrategy"
+        render={({ field }) => (
+          <FormItem>
+            <Field.Header
+              label="Machine uniqueness strategy"
+              variant="stacking"
+              tooltip={PolicyAttributeDescriptions.machineUniquenessStrategy}
```
ezekg, line 94: Can we mute or hide the "i" icon unless the attribute name is hovered? I think it's a bit visually overwhelming:

Image

Also, when breaking across multiple lines, the icon should inline be at the end of the line.

src/components/policies/fields/node-locked.tsx
```diff
@@ -0,0 +1,568 @@
...
+                <FormControl>
+                  <SelectTrigger className="w-full">
+                    <SelectValue placeholder="Select one..." />
+                  </SelectTrigger>
+                </FormControl>
+                <SelectContent>
+                  {Object.values(OverageStrategy).map(
+                    (basis: OverageStrategy) => (
+                      <SelectItem key={basis} value={basis}>
+                        {humanize(String(basis))}
```
ezekg: The humanize function shouldn't remove the decimal from floating point numbers, e.g. this currently reads like 5x/25x overages depending on if you ignore the first number:

Image

tbh, I'd rather not use humanize and just hard code a code-to-names mapping instead — that way we're in full control of the presented text, e..g "Unique Per-Policy" vs "Unique Per Policy."

src/components/policies/fields/node-locked.tsx
```diff
@@ -0,0 +1,568 @@
...
+              >
+                <FormControl>
+                  <SelectTrigger className="w-full">
+                    <SelectValue placeholder="Select one..." />
+                  </SelectTrigger>
+                </FormControl>
+                <SelectContent>
+                  {Object.values(OverageStrategy).map(
+                    (basis: OverageStrategy) => (
+                      <SelectItem key={basis} value={basis}>
```
ezekg: Nit: this var is `basis`, but should be `strategy`.

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,825 @@
...
+      if (step > 0) {
+        setCompletedStep((prev) => {
+          const next = new Set(prev)
+          next.add(steps[step].key)
+
+          return next
+        })
+      }
+
+      goTo(step + 1)
```
ezekg, line 177: Not sure where it happens, but on "next step," we should only focus the name field if it hasn't been set yet, otherwise focus the next field or nothing at all.

src/components/policies/fields/requirements.tsx
```diff
@@ -0,0 +1,482 @@
...
+            )}
+          />
+        </div>
+      </div>
+
+      {includeMeta && (
+        <div className="mt-6">
+          <FormField
+            control={form.control}
+            name="metadata"
```
ezekg, line 457: It won't let me submit the form without metadata, which shouldn't be required:

Image

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,825 @@
...
+        <Dialog open={isAttributesOpen} onOpenChange={onClose}>
+          <DialogContent
+            disableOverlay
+            className="min-w-fit"
+            onWheelCapture={(e) => e.stopPropagation()}
+            onTouchMoveCapture={(e) => e.stopPropagation()}
+          >
+            <DialogHeader className="h-fit items-start border-b border-accent p-2">
+              <DialogDescription>
+                <BadgeGroup prefix="Creating a" suffix="policy">
```
ezekg, line 317: I see a markup error saying that a `div` can't be inside a `p` tag. May need to use a `span` instead:

Image

src/components/policies/create/scratch-form.tsx
```diff
@@ -0,0 +1,315 @@
...
+        title: "Scope",
+        fields: [
+          "requireProductScope",
+          "requirePolicyScope",
+          "requireMachineScope",
+          "requireFingerprintScope",
+          "requireComponentsScope",
+          "requireUserScope",
+          "requireChecksumScope",
+          "requireVersionScope",
```
ezekg, line 129: Not sure I like the styling of the checkboxes here:

Image

Maybe we can try left/right-aligning the checkbox so that it looks more natural?

Image

Image

Image

src/components/policies/create/scratch-form.tsx
```diff
@@ -0,0 +1,315 @@
...
+      if (!ok) return
+    }
+
+    if (!last) {
+      setCompleted((prev) => new Set(prev).add(current.key))
+      goTo(step + 1)
+    } else {
+      const payload = form.getValues()
+      onCreate(payload)
+      onClose()
```
ezekg, line 161: Should close the create modal as well, then redirect to the new policy.

ezekg (changes requested):
Overall, looks really great. Way better than the current flow. Have a few things I'd like to get buttoned up before merging. I may also eventually go in and reorganize things in the scratch flow, but that'll be in a separate PR once I can figure out the best categorization.

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,825 @@
...
+import * as Policies from "@/components/policies"
+import StepProgress from "@/components/step-progress"
+import DocumentationLink from "@/components/documentation-link"
+import CollapsedBreadcrumb from "@/components/collapsed-breadcrumb"
+
+import ScratchForm from "./scratch-form"
+import ParametersForm, { ParametersValues } from "./parameters-form"
+
+export const BaseSchema: z.ZodType<PolicyFormValues> = z
+  .object({
```
ezekg: Oh yeah, before I forget — all policies are required to have a product relationship. You would've found this out eventually when integrating with the API, but figured I'd note it now.

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,825 @@
...
+              name: z.string().min(1),
+              code: z.string().min(1),
+              metadata: z.record(z.string()).optional(),
+            }),
+          )
+          .optional(),
+      })
+      .optional(),
+
+    metadata: z.record(z.string()).default({}),
```
ezekg: Does `z.record` support nested objects? The API technically allows 2 levels of nesting, so would eventually like a way for users to write "custom" JSON themselves, instead of the key-value input fields. Like some sort of "advanced" metadata input. That can ofc be for later, but just wanted to note it here before I forgot again.

src/components/policies/create/scratch-form.tsx
ezekg: I say make the scratch form checkboxes look like what you have in the wizard form:

image

src/components/policies/fields/requirements.tsx
```diff
@@ -0,0 +1,482 @@
...
+                  )
+                }
+              >
+                <FormControl>
+                  <SelectTrigger className="w-full">
+                    <SelectValue />
+                  </SelectTrigger>
+                </FormControl>
+                <SelectContent>
+                  {Object.values(CheckInInterval).map(
```
ezekg: Currently, if a check-in interval is chosen, there's no way to unset it:

image

Maybe add a null field to all select inputs?

src/components/policies/create/modal.tsx
ezekg: Also need to be sure that numerics and booleans are allowed and handled correctly, i.e. not all values will be strings. The current UI does this implicitly, e.g. if it looks like an integer, it's parsed as such. But this ends up being problematic when a user e.g. wants a numeric string. Instead, I think the metadata input should have a value type selector, which explicitly parses the value as a specific type, and errors if unable.

This type selector could also be expanded to have support for a nested structure in the future, e.g. object/array value types, though I don't think that should be a priority right now.

I think these values types should be explicitly supported now:

- `string` (default)
- `boolean`
- `integer`
- `float`

Any blank values should be parsed as `null`.

ezekg: Btw, fine and preferred if we address this in in a separate PR. I can move this to an issue if you'd like, or you can (I can't remember where you're currently keeping issues/backlog/tickets).

image

src/components/policies/create/modal.tsx
ezekg: Should probably add the product relationship in an initial step alongside name.

src/components/policies/fields/requirements.tsx
ezekg: Related to the comment https://github.com/keygen-sh/keygen-portal/pull/16#discussion_r2391996846, but I think it'd look better visually if the checkboxes were aligned to the right side of the sibling dropdowns/value container area, instead of the left side.

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,825 @@
...
+    requireMachineScope: z.boolean().optional(),
+    requireFingerprintScope: z.boolean().optional(),
+    requireComponentsScope: z.boolean().optional(),
+    requireUserScope: z.boolean().optional(),
+    requireChecksumScope: z.boolean().optional(),
+    requireVersionScope: z.boolean().optional(),
+
+    entitlements: z
+      .object({
+        link: z.array(z.string()).optional(),
```
ezekg: I think using "link" here is confusing, because link is already an established term in the API, especially in the context of relationships. I'd rather see attach here, since that's the action taken for existing entitlements.

Or, if you used a union instead, it may be easier to simplify with nested entitlements later on after keygen-sh/keygen-api#932 is eventually implemented:

```ts
// policy schema
z.object({
  // …
  entitlements: z.array(
    z.union([
      z.object({id: z.string()}).strict(),
      z.object({name: z.string(), code: z.string(), metadata: z.record(z.unknown())}).strict(),
    ]),
  ),
  // …
})

// partition (just spit-balling some pseudo-code)
const {toAttach, toCreate} = params.entitlements.reduce((acc, entitlement) => {
  if ("id" in entitlement) {
    acc.toAttach.push(entitlement)
  } else {
    acc.toCreate.push(entitlement)
  }
  return acc
}, {
  toAttach: [],
  toCreate: [],
})

// create entitlements before policy (what we'd do now)
for (let {name, code, metadata} of toCreate) {
  const entitlement = await api.entitlements.create({name, code, metadata})

  toAttach.push({id: entitlement.id})
}

const policy = await api.policies.create({entitlements: toAttach})

// vs eventually what we'd do later…
const policy = await api.policies.create({entitlements: params.entitlements})
```

src/components/policies/fields/timed.tsx
Cazden: This is disabled because I'm not sure what the exact values for Renewal Strategy are since it's a WIP feature (auto-renewals/subscription-based). I used a placeholder in the type as well (`AUTO_RENEW`), and had planned on updating this once the feature was available. I still included the field so it's ready and just needs to be enabled. I should've left a comment, and I can add one unless you want to go ahead and enable the button now?

src/components/policies/fields/node-locked.tsx
Cazden: I'd had that on my list of todos and was probably going to expand the humanize function in another PR and retroactively fix this particular case. I hadn't thought of cases like "Unique Per-Policy" though, which would probably require a mapping anyway, so I agree hardcoding would be the better solution here so it's consistent across all attributes.

Just now pushed a change for this in [9de245f](https://github.com/keygen-sh/keygen-portal/pull/16/commits/9de245fc93c2bd19fd1b09c72cb1eac9fb77e04d) so there's a map for attribute options now, but `humanize` is still being used in a couple of files for humanizing attribute labels (not the options but attributes themselves, i.e. "MAX_MACHINES" > "Max machines"). I'm not sure if creating a map for the attribute labels is necessary but thought I'd bring it up if you wanted to get rid of `humanize` altogether.

Fixed:
image

src/components/policies/fields/node-locked.tsx
Cazden: Addressed hiding the icon in [f6975aa](https://github.com/keygen-sh/keygen-portal/pull/16/commits/f6975aafa80a4d052af113ea0578264a9855d3ec), so it'll reveal after hovering the attribute label now.

Hovering label:
image

Hovering icon:
image

---

Making the icon inline still needs to be fixed, had that one on my todos for a bit. I'll leave this open until that's resolved.

src/components/policies/create/scratch-form.tsx
Cazden: Agree completely. I probably spent the least amount of time on this form, and was going to either refactor inline or maybe look at doing a custom toggle switch input (yes/no) just to give the UI some life.

Refactored to inline in [8bfb939](https://github.com/keygen-sh/keygen-portal/pull/16/commits/8bfbf939f35d6579b7af4ad4640daa7798fdc22c).

src/components/policies/create/modal.tsx
Cazden: > Does `z.record` support nested objects? The API technically allows 2 levels of nesting, so would eventually like a way for users to write "custom" JSON themselves, instead of the key-value input fields. Like some sort of "advanced" metadata input. ...

I think that's an awesome idea. Out of the box it doesn't support nesting, but you can add it with a union (for 1-2 levels of nesting), e.g.
```
const Metadata = z.union([z.string(), z.record(z.string())])

const Schema = z.object({
  metadata: z.record(Metadata).default({}),
})

// So this would be valid:
{
  metadata: {
    foo: "bar",
    nested: {
      hello: "world"
    }
  }
}
```

That's simplified because it's just handling strings, but you get the idea. If you wanna go beyond 2 levels, `z.lazy` can make it recursive, e.g. `const Metadata = z.lazy(() => z.union([z.string(), z.record(Metadata)]))`

---
> Also need to be sure that numerics and booleans are allowed and handled correctly, i.e. not all values will be strings. ...
> Btw, fine and preferred if we address this in in a separate PR. ...

Sounds good, I've moved this to an [issue](https://github.com/keygen-sh/keygen-portal/issues/17) and we can handle it in another PR.

src/components/policies/fields/timed.tsx
ezekg: Ah, right! You can just remove renewal strategy for now and we'll implement later.

src/components/policies/create/modal.tsx
Cazden: Added in [354a7d9](https://github.com/keygen-sh/keygen-portal/pull/16/commits/354a7d9fefd65efb7e9e9ef0b625e601fddea26b)—I've also separated the `name` field into a new general form step that includes the product relationship.

src/components/policies/fields/node-locked.tsx
Cazden: I'm going to mark this as resolved for now but feel free to open it back up if you want to touch on the above.

src/components/policies/fields/requirements.tsx
Cazden: Here's a "Clear" option I've added to all the Select fields across the form that will unset the selected value:

image

src/components/field/header.tsx
```diff
@@ -34,6 +34,14 @@ export default function FieldHeader({
 }: FieldHeaderProps) {
   const isMobile = useMobile()
 
+  function splitLastWord(s: string): { head: string; tail: string } {
```
ezekg: For performance reasons, this should not be an inline function, because it's redefined every render. Pull it out into a util or even just a component-level function.

src/components/policies/create/modal.tsx
```diff
@@ -180,6 +173,9 @@ export const BaseSchema: z.ZodType<PolicyFormValues> = z
           .optional(),
       })
       .optional(),
+    product: z.object({
+      attach: z.string().min(1, "Product is required."),
```
ezekg: Not sure "attach" makes sense here. Maybe just make `product` be a string ID, or at the very least rename this to `product.id` so that it makes sense? Attach is a very specific one-to-many action for entitlements and other resources like users. And also just to be clear: we'll never support creating a new product inline during policy creation like we will (eventually) for the entitlements relationship, in case you were designing around that possibility.

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,1103 @@
...
+        metered: values.metered ?? [],
+        advanced: !!values.advanced,
+        offline: !!values.offline,
+      }
+
+      setSelection(newSelection)
+      setCompletedStep(new Set())
+
+      form.reset(
+        { name: "", ...getFormDefaults(newSelection) },
```
ezekg: If we used composed schemas (as discussed on Discord), we could use `{name: '', ...schema.parse({})}` to fill in schema defaults. (If you do this, I'd recommend creating a `getSchemaDefaults` util or something so that intent is clear, since using `parse` in this way to obtain the schema's default is not obvious even though it works.)

ezekg (commented):
First pass review. Still going through things, but figured I'd submit what I have so that you can keep some of the comments in mind during the Zod schema refactor.

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,1103 @@
...
+          No attributes required.
+        </div>
+      ),
+    })
+  }
+
+  return steps
+}
+
+export function getFormDefaults(selection?: PolicyParameterSelection) {
```
ezekg: When possible, explicitly type a function's return value.

src/components/policies/create/parameters-form.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+export type ParametersValues = z.infer<typeof schema>
+
+type ParametersFormProps = {
+  initial?: Partial<PolicyParameterSelection>
+  onSubmit: (values: ParametersValues) => void
+  onStartScratch?: () => void
+}
+
+export default function ParametersForm({
```
ezekg: Nit: I think this would be clearer if it was named `TemplateForm`, i.e. use "template" verbiage instead of "parameters," since we're applying predetermined license model templates to the new policy. That's how I've always thought of this new creation flow, so "template" makes the most sense imo.

src/pages/app/policy/details.tsx
```diff
@@ -0,0 +1,680 @@
...
+                        navigate({
+                          to: "/$id/app/products/$productId",
+                          params: {
+                            id: keygen.config.id,
+                            productId: product.id,
+                          },
+                        })
+                      }}
+                    >
+                      {product?.attributes.name} ({productId?.slice(0, 8)})
```
ezekg: We don't need to include the product ID here — just the name is fine (it looks cleaner).

src/types/policies.ts
```diff
@@ -0,0 +1,547 @@
...
+  KEEP_DEAD = "KEEP_DEAD",
+}
+export enum HeartbeatResurrectionStrategy {
+  NO_REVIVE = "NO_REVIVE",
+  "1_MINUTE_REVIVE" = "1_MINUTE_REVIVE",
+  "2_MINUTE_REVIVE" = "2_MINUTE_REVIVE",
+  "5_MINUTE_REVIVE" = "5_MINUTE_REVIVE",
+  "10_MINUTE_REVIVE" = "10_MINUTE_REVIVE",
+  "15_MINUTE_REVIVE" = "15_MINUTE_REVIVE",
+  "ALWAYS_REVIVE" = "ALWAYS_REVIVE",
```
ezekg: Nit: unneeded quotes here for `ALWAYS_REVIVE`.

src/types/policies.ts
```diff
@@ -0,0 +1,547 @@
...
+  DEACTIVATE_DEAD = "DEACTIVATE_DEAD",
+  KEEP_DEAD = "KEEP_DEAD",
+}
+export enum HeartbeatResurrectionStrategy {
+  NO_REVIVE = "NO_REVIVE",
+  "1_MINUTE_REVIVE" = "1_MINUTE_REVIVE",
+  "2_MINUTE_REVIVE" = "2_MINUTE_REVIVE",
+  "5_MINUTE_REVIVE" = "5_MINUTE_REVIVE",
+  "10_MINUTE_REVIVE" = "10_MINUTE_REVIVE",
+  "15_MINUTE_REVIVE" = "15_MINUTE_REVIVE",
```
ezekg: Maybe just use e.g. `FIFTEEN_MINUTE_REVIVE` instead of the awkward (DX-wise) numeric key prefix?

src/types/policies.ts
```diff
@@ -0,0 +1,547 @@
...
+    [HeartbeatCullStrategy.DEACTIVATE_DEAD]: "Deactivate Dead",
+    [HeartbeatCullStrategy.KEEP_DEAD]: "Keep Dead",
+  },
+  heartbeatResurrectionStrategy: {
+    [HeartbeatResurrectionStrategy.NO_REVIVE]: "No Revive",
+    [HeartbeatResurrectionStrategy["1_MINUTE_REVIVE"]]: "1 Minute Revive",
+    [HeartbeatResurrectionStrategy["2_MINUTE_REVIVE"]]: "2 Minute Revive",
+    [HeartbeatResurrectionStrategy["5_MINUTE_REVIVE"]]: "5 Minute Revive",
+    [HeartbeatResurrectionStrategy["10_MINUTE_REVIVE"]]: "10 Minute Revive",
+    [HeartbeatResurrectionStrategy["15_MINUTE_REVIVE"]]: "15 Minute Revive",
```
ezekg: This is what I mean by "awkward" here: https://github.com/keygen-sh/keygen-portal/pull/16/files#r2414324169

Cazden: Yeah I wasn't a fan of the usage either, in my head I was thinking it's better to be consistent with the API so I opted for the string keys. I think `FIFTEEN_MINUTE_REVIVE` is much cleaner though.

src/components/policies/fields/node-locked.tsx
ezekg: There are still some oddities for attribute labels too, like "Check In …" instead of "Check-in …", but that's probably too nitpicky (because the former still makes enough sense i.e. it's not like it's grammatically incorrect or anything) so I'm fine leaving it as-is for now. We can always go back and adjust later on in a separate PR.

src/components/policies/fields/lease-based.tsx
```diff
@@ -0,0 +1,385 @@
...
+        name="heartbeatDuration"
+        render={({ field }) => (
+          <FormItem>
+            <Field.Header
+              label="Heartbeat duration"
+              variant="stacking"
+              tooltip={PolicyAttributeDescriptions.heartbeatDuration}
+            >
+              <FormControl>
+                <Input
```
ezekg: Could we make this a duration field like `policy.duration`? It will need to support smaller units e.g. seconds/minutes/hours.

src/components/policies/fields/lease-based.tsx
```diff
@@ -0,0 +1,385 @@
...
+            control={form.control}
+            name="heartbeatDuration"
+            render={({ field }) => (
+              <FormItem>
+                <Field.Header
+                  label="Heartbeat duration"
+                  tooltip={PolicyAttributeDescriptions.heartbeatDuration}
+                >
+                  <FormControl>
+                    <Input
```
ezekg: Same as https://github.com/keygen-sh/keygen-portal/pull/16/files#r2426852759

ezekg (dismissed):
Love it. I have a few minor changes, but other than that LGTM!! 🤘

src/components/policies/create/modal.tsx
```diff
@@ -0,0 +1,731 @@
...
+            disableOverlay
+            className="min-w-fit"
+            onWheelCapture={(e) => e.stopPropagation()}
+            onTouchMoveCapture={(e) => e.stopPropagation()}
+          >
+            <DialogHeader className="h-fit items-start border-b border-accent p-4">
+              <DialogTitle className="mb-1 text-base">New policy</DialogTitle>
+              <DialogDescription>
+                <span className="text-sm">
+                  Select all parameters that apply to your license
```
ezekg:
```suggestion
                  Select all parameters that apply to the policy
```

src/lib/utils.ts
```diff
@@ -53,3 +53,64 @@ export function diff<T extends object>(
...
+  const weeks = Math.trunc(seconds / WEEK)
+  seconds -= weeks * WEEK
+  const days = Math.trunc(seconds / DAY)
+  seconds -= days * DAY
+  const hours = Math.trunc(seconds / HOUR)
+  seconds -= hours * HOUR
+  const minute = Math.trunc(seconds / MINUTE)
+  seconds -= minute * MINUTE
+
+  return { years, months, weeks, days, hours, minute, seconds }
```
ezekg:
```suggestion
  return { years, months, weeks, days, hours, minutes, seconds }
```

This is causing minutely durations to not render because `date-fns` expects a `minutes` key:

Image

ezekg (approved):

PR #19, opened 2025-11-03
add edit policies interface

Cazden:
Adds interface for updating existing policies along with mock update functionality.

### DIFF
--- a/src/pages/app/policy/details.tsx
+++ b/src/pages/app/policy/details.tsx
@@ -105,6 +105,7 @@ export default function PolicyDetails() {
 
   const isMobile = useMobile()
   const [open, setOpen] = useState({
+    edit: false,
     delete: false,
     attributes: false,
   })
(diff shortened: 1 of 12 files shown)


src/pages/app/policy/details.tsx
```diff
@@ -123,7 +124,9 @@ export default function PolicyDetails() {
   }, [])
 
   const handleOpen = (key: keyof typeof open) => {
+  // FIXME(cazden) Opening/closing dialogs has performance issues
     setOpen({
```
ezekg, line 128: This can be simplified to:

```ts
setOpen({
  ...open,
  [key]: !open[key],
})
```

src/pages/app/policy/details.tsx
```diff
@@ -123,7 +124,9 @@ export default function PolicyDetails() {
   }, [])
 
   const handleOpen = (key: keyof typeof open) => {
```
ezekg: I think this should be renamed to `toggleOpen` since it also closes if already open.

ezekg (dismissed):
Two minor comments, aside from the sporadic performance issue we've previously discussed.

Feel free to merge and tackle all that in a separate PR.

ezekg (approved):
LGTM.

PR #20, opened 2025-11-06
fix policy dialog performance

Cazden:
Fixes the studder/performance issues when opening and closing dialogs on the policy details page.

### DIFF
--- a/src/components/policies/fields/all.tsx
+++ b/src/components/policies/fields/all.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from "react"
+import { useState, useEffect } from "react"
 import { useFormContext, useFieldArray, useWatch } from "react-hook-form"
 
 import { Input } from "@/components/ui/input"
@@ -72,8 +72,13 @@ export default function AllFields({
     name: "entitlements.create" as const,
   })
 
-  const duration = useWatch({ control: form.control, name: "duration" })
+  const [mounted, setMounted] = useState(false)
+  useEffect(() => {
+    const id = requestAnimationFrame(() => setMounted(true))
+    return () => cancelAnimationFrame(id)
+  }, [])
 
+  const duration = useWatch({ control: form.control, name: "duration" })
   useEffect(() => {
     if (duration !== null) return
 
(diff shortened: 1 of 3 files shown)


ezekg (dismissed):
I can't reproduce the state/performance issue after this, so seems to have resolved it. Though there is still a short delay between clicking edit and the modal actually appearing, though that might be the animation? Either way, that can be addressed later.

Cazden:
Fixed the performance of updating/closing by refactoring state/form rendering and adding a skeleton UI to the form fields component. This should conclude any performance issues we were having with the dialog.

src/components/policies/fields/all.tsx
```diff
@@ -147,6 +150,71 @@ export default function AllFields({
...
   }, [requireHeartbeat])
 
+  if (!mounted || productsLoading) {
+    return (
+      <div className={cn("p-4", className)}>
+        <h2 className="text-content-loud/90">General</h2>
+        <section className="mt-4 flex flex-col md:flex-row">
+          <div className="w-full space-y-3">
+            <Skeleton className="h-5 w-32" />
+            <Skeleton className="h-9 w-3/5" />
```
ezekg, line 160: Not sure I like this at first glance. Seems like each skeleton should be a part of each field, displayed during load, and not maintained separately. This way seems like it'll become a maintenance issue as we add/change fields, since it'll easily become out of sync. Is there a particular reason you did it this way?

The whole mount `useEffect` seems like a code smell imo.

Cazden: Skeleton isn't a 1:1 for every field, it's just a generalization:
image

I was thinking separating the two would be better and make scanning the form fields visually easier without a bunch of conditionals and extra JSX nearby, and it's generic enough that we wouldn't necessarily need to refactor it if we made any field changes. I can refactor if you'd prefer to have it with the fields though.

---

As far as the `useEffect`, this is here because rendering all these form fields has a high performance toll and adding a frame delay allows the dialog to open smoothly without getting bogged down from loading all of the fields.

ezekg: Makes sense. 👍

ezekg (approved):
LGTM.

PR #22, opened 2025-11-13
add entitlements api integration

Cazden:
Adds CRUD for entitlements and replaces mock entitlements data with API queries.


src/components/entitlements/create/modal.tsx
```diff
@@ -41,38 +40,26 @@ export default function EntitlementsCreateModal({
...
+          toast({ message: "Entitlement created", variant: "success" })
+          onSelectEntitlement(entitlement)
+          onClose()
+        },
+        onError: (error) => {
+          if (
+            typeof error === "object" &&
+            error &&
+            "code" in error &&
+            error.code === EntitlementErrorCode.CODE_TAKEN
```
ezekg: This seems overly verbose. Why not just type the param  `error: Error` and then `error?.code ===`? Or even better, assuming that error will always be non-null seems to be fine? In what cases would this be null/undefined? 

(Assuming this is for [Tanstack Query](https://tanstack.com/query/latest/docs/framework/react/reference/useMutation#:~:text=onError:%20\(err:) ofc.)

src/components/policies/create/modal.tsx
```diff
@@ -225,22 +250,93 @@ export default function PoliciesCreateModal({
...
+        })
+
+        if (isScratchOpen) {
+          // Throw to render error messages in scratch form
+          throw {
+            kind: "entitlements-validation",
+            fieldErrors,
+            nextAttach,
+            nextCreate,
+          }
```
ezekg: I would lean towards the opinion that throwing non-`Error` objects is non-idiomatic JS, but I'm not a JS expert so I could be wrong. To me, this seems like it should be a custom explicitly-named `Error` class that exposes these fields, e.g. `class CreateEntitlementValidationError extends Error { ... }`, then you can switch on `typeof err === CreateEntitlementValidationError` in the policy scratch form to handle appropriately.

src/components/policies/create/modal.tsx
```diff
@@ -225,22 +250,93 @@ export default function PoliciesCreateModal({
...
+        const nextAttach = Array.from(
+          new Set([...attachIds, ...createdEntitlements.map((e) => e.id)]),
+        )
+
+        const nextCreate = failed.map((index) => toCreate[index])
+
+        const fieldErrors: {
+          path: `entitlements.create.${number}.code`
+          message: string
+        }[] = failed.map((index, newIndex) => {
```
ezekg: If this is an error shape you plan on using for other validations, I'd add an explicit type/class/interface for it so that you don't have to write this inline type everywhere, e.g. `FormFieldError` or whatever.

src/components/policies/create/scratch-form.tsx
```diff
@@ -157,10 +157,35 @@ export default function PoliciesScratchForm({
...
+          if (nextAttach) {
+            form.setValue("entitlements.attach", nextAttach)
+          }
+          if (nextCreate) {
+            form.setValue("entitlements.create", nextCreate)
+          }
+
+          fieldErrors.forEach(
+            (fieldError: {
+              path: `entitlements.create.${number}.code`
```
ezekg: TIL you can type-check the internals of string literals!

src/components/policies/edit/edit-form.tsx
```diff
@@ -18,31 +23,110 @@ import {
...
+          }),
+        ),
+      )
+
+      const succeeded = entitlementsResults
+        .map((result, index) => (result.status === "fulfilled" ? index : -1))
+        .filter((index) => index !== -1)
+      const failed = entitlementsResults
+        .map((result, index) => (result.status === "rejected" ? index : -1))
+        .filter((index) => index !== -1)
```
ezekg: Nit: instead of the map-filter, you could partition using reduce:

```ts
const { succeeded, failed } = entitlementResults.reduce(
  (results, result, index) => {
    if (result.status === "fulfilled") {
      results.succeeded.push(result.value) // or index if you want indexes
    } else {
      results.failed.push(result.value)
    }
    return results
  },
  {
    succeeded: [],
    failed: [],
  },
)
```

Extracting into a partitioning util might also be useful if this is a pattern that'll be used elsewhere:

```ts
type Partition<T> = [T[], T[]]

function partition<T>(values: T[], predicateFn: (v: T) => boolean): Partition<T> {
  return values.reduce(
    ([x, y], v): Partition<T> => (predicateFn(v) ? [[...x, v], y] : [x, [...y, v]]),
    [[], []],
  )
}
```

Then use like:

```ts
const [succeeded, failed] = partition(results, (result: PromiseSettledResult) =>
  result.status === 'fulfilled'
)

const entitlements = succeeded.map(r => r.value)
const errors = failed.map(r => r.reason)
```

Again, if it's a consistent pattern, you could even add a wrapper util e.g. `partitionFulfilled`.

src/components/policies/edit/edit-form.tsx
```diff
@@ -18,31 +23,110 @@ import {
...
+        form.setValue("entitlements.attach", nextAttach)
+        form.setValue("entitlements.create", nextCreate)
+
+        failed.forEach((index, newIndex) => {
+          const result = entitlementsResults[index]
+
+          let message = ""
+          if (
+            result.status === "rejected" &&
+            result.reason?.code === EntitlementErrorCode.CODE_TAKEN
```
ezekg: Again, refactoring via partitioning and better types might clean this up a bit. Right now you're dealing with raw `PromiseSettledResult` primitives, which isn't exactly great to work with because you have to assert status and field existence before working with it. Imo, it'd be worth refactoring into more explicit wrapper types with a cleaner set of fields exposed that can be reused across CRUD resources.

ezekg (changes requested):
Looking great. UI/UX itself is fine, but have a few refactors I'd like to see.

src/components/policies/create/modal.tsx
ezekg: Might be worth an ESLint rule tbh: https://eslint.org/docs/latest/rules/no-throw-literal

src/components/entitlements/create/modal.tsx
Cazden: You're absolutely right! My apologies. Fixed in [#cb31](https://github.com/keygen-sh/keygen-portal/pull/22/commits/cb31bd5b73a931d6fcd2d6859c750993f8d27b3a).

src/components/policies/create/modal.tsx
Cazden: Yeah, I think a little more type-safety was warranted here. Both requested changes have been addressed in 08a387b07b89cd5da0dc0ba1b3c47d08fb1538c6 and fb844fff1b943f46d4a965e8dd796d5d691f382d, along with the `FormFieldError` you suggested in another [comment](https://github.com/keygen-sh/keygen-portal/pull/22#discussion_r2535518817).

src/components/entitlements/create/modal.tsx
ezekg: > You're absolutely right!

👁️👄👁️

src/components/policies/create/modal.tsx
```diff
@@ -259,54 +260,34 @@ export default function PoliciesCreateModal({
...
-          },
-        )
+      const nextAttach = [...entitlementIds]
+      const nextCreate = errors.map(({ index }) => toCreate[index])
+
+      form.setValue("entitlements.attach", nextAttach)
+      form.setValue("entitlements.create", nextCreate)
+
+      if (errors.length > 0) {
+        const fieldErrors: FormFieldError<PolicyFormValues>[] = []
```
ezekg: This variable seems unused. It's always empty.

ezekg (changes requested):
Looks good. One comment on a regression in the create modal due to the refactor.

src/components/policies/create/modal.tsx
Cazden: Missed that! Fixed in 05e925af0e6684cd2203db6516681a085d0c2f1f.

ezekg (dismissed):
Rad. LGTM.

src/components/policies/create/modal.tsx
```diff
@@ -274,20 +274,21 @@ export default function PoliciesCreateModal({
...
-
-          form.setError(`entitlements.create.${index}.code`, {
-            type: "validate",
-            message,
-          })
-        })
+        const fieldErrors: FormFieldError<PolicyFormValues>[] = errors.map(
+          (error, index) => {
+            let message = ""
+            if (error.reason?.code === EntitlementErrorCode.CODE_TAKEN) {
```
ezekg: Nit: I don't think reason will ever be null here.

ezekg (approved):

PR #23, opened 2025-11-26
Switch from npm to pnpm

ezekg:
Replaces npm with pnpm to improve security posture. Also does some house keeping e.g. adds basic README, LICENSE, and a SECURITY policy. Will try to add a basic CI workflow later today as well, e.g. install/build/lint/etc., until we have a test suite.


Cazden (approved):
LGTM!

PR #25, opened 2025-12-01
Add table column helper for resources

ezekg:
Fixes the `any` type lint errors in our Tanstack Table usage via introducing a resource-aware column helper, i.e. `helper.id()` for accessing the ID, `helper.attr()` for accessing attributes, and `helper.rel()` for accessing relationships. This new helper correctly infers types according to the accessed resource field.

### DIFF
--- a/src/components/environments/view/list.tsx
+++ b/src/components/environments/view/list.tsx
@@ -1,6 +1,6 @@
 import { useMemo } from "react"
-import { ColumnDef, createColumnHelper } from "@tanstack/react-table"
 
+import { createResourceColumnHelper } from "@/lib/tables"
 import { Environment } from "@/types/environments"
 
 import { useListEnvironments } from "@/queries/environments"
@@ -17,20 +17,18 @@ export default function EnvironmentsList({
 }: EnvironmentsListProps) {
   const { data: environments = [], isLoading } = useListEnvironments()
 
-  const column = createColumnHelper<Environment>()
-  const columns = useMemo<ColumnDef<Environment, any>[]>(
+  const column = createResourceColumnHelper<Environment>()
+  const columns = useMemo(
     () => [
-      column.accessor((row) => row.attributes.name, {
+      column.id( { header: "ID" }),
+      column.attr("name", {
         header: "Name",
-        id: "attributes.name",
       }),
-      column.accessor((row) => row.attributes.code, {
+      column.attr("code", {
         header: "Code",
-        id: "attributes.code",
       }),
-      column.accessor((row) => row.attributes.isolationStrategy, {
+      column.attr("isolationStrategy", {
         header: "Isolation Strategy",
-        id: "attributes.isolationStrategy",
       }),
     ],
     [],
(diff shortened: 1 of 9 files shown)


src/components/environments/view/list.tsx
```diff
@@ -17,20 +17,18 @@ export default function EnvironmentsList({
 }: EnvironmentsListProps) {
   const { data: environments = [], isLoading } = useListEnvironments()
 
-  const column = createColumnHelper<Environment>()
-  const columns = useMemo<ColumnDef<Environment, any>[]>(
+  const column = createTableColumnHelper<Environment>()
+  const columns = useMemo(
```
ezekg, line 21: imo, it's fine to let TS infer the type here because it can get super complex if all columns aren't strings.

src/lib/tables.ts
```diff
@@ -0,0 +1,38 @@
+import { ColumnDef, createColumnHelper } from "@tanstack/react-table"
+
+interface TableResource {
```
ezekg, line 3: This is simply the minimal representation of what our table column helper needs in order to function.

src/pages/app/entitlement/list.tsx
```diff
@@ -24,33 +24,29 @@ export default function PoliciesList() {
 
   const [open, setOpen] = useState(false)
 
-  const column = createColumnHelper<Entitlement>()
-  const columns = useMemo<ColumnDef<Entitlement, any>[]>(
+  // TODO(ezekg) extract all this out into e.g. useEntitlementTableColumns so it can be used elsewhere?
+  const column = createTableColumnHelper<Entitlement>()
```
ezekg, line 28: Probably shouldn't be recreating the helper on every render, so might be better to hoist outside of the component (since it seems to be stateless), or put into some sort of memoized use hook/function.

src/pages/app/policy/list.tsx
```diff
@@ -31,44 +31,38 @@ export default function PoliciesList() {
...
-          cell: (info) =>
-            info.getValue() || (
+      column.rel("product", {
+        sortingFn: "alphanumeric",
+        header: "Product",
+        cell: (info) => {
+          const productId = info.getValue()?.data?.id ?? ""
+
+          return (
+            productNameById.get(productId) ?? (
```
ezekg, line 51: Not sure if fetching makes sense in `cell`, but to me it seemed like keeping the `row => row.foo.bar` accessor pure, with the display functions being where the fetching/formatting happens so that e.g. loading can properly be handled via suspense, a spinner, a skeleton, etc.

ezekg: I guess fetching is actually handled by an effect, and passed into memo as a dependency, so disregard the above. This seems fine to me, but lmk if it seems odd to you.

src/pages/app/entitlement/list.tsx
Cazden: Good idea. I think using a hook would cleanup repeated component logic. As discussed in discord I'll open another PR for this.

Cazden (approved):
LGTM!

PR #26, opened 2025-12-01
Refactor enum naming from uppercase to pascalcase

ezekg:
I misspoke awhile back saying we should use uppercase enums fields when [the TS convention is pascalcase](https://www.typescriptlang.org/docs/handbook/enums.html).

This has been bugging me, so I figured I'd open a PR to fix it.


Cazden (changes requested):
Looks great, just missing one enum here:
https://github.com/keygen-sh/keygen-portal/blob/9fbfbb871459e209082293984f1912cadfd130b7/src/components/sidebar/panel.tsx#L55-L65

Cazden (approved):
LGTM!

PR #28, opened 2025-12-02
Refactor list components table columns into hooks

Cazden:
Referenced in https://github.com/keygen-sh/keygen-portal/pull/25#discussion_r2578869399.

Refactors each resource's list table columns with resource-specific hooks to abstract column logic and improve performance via memoization.

### DIFF
--- a/src/hooks/use-policy-table-columns.tsx
+++ b/src/hooks/use-policy-table-columns.tsx
@@ -2,23 +2,14 @@ import { useMemo } from "react"
 
 import { Policy } from "@/types/policies"
 
-import { useListProducts } from "@/queries/products"
-
 import { createTableColumnHelper } from "@/lib/tables"
 
 import ClipboardButton from "@/components/clipboard-button"
+import ProductCell from "@/components/tables/product-cell"
 
 const column = createTableColumnHelper<Policy>()
 
 export function usePolicyTableColumns() {
-  const { data: products = [] } = useListProducts()
-
-  const productNameById = useMemo(() => {
-    const map = new Map<string, string>()
-    for (const product of products) map.set(product.id, product.attributes.name)
-    return map
-  }, [products])
-
   const columns = useMemo(
     () => [
       column.id({
@@ -31,15 +22,7 @@ export function usePolicyTableColumns() {
       column.rel("product", {
         sortingFn: "alphanumeric",
         header: "Product",
-        cell: (info) => {
-          const productId = info.getValue()?.data?.id ?? ""
-
-          return (
-            productNameById.get(productId) ?? (
-              <span className="text-muted-foreground">--</span>
-            )
-          )
-        },
+        cell: (info) => <ProductCell id={info.getValue()?.data?.id} />,
       }),
       column.attr("created", {
         sortingFn: "datetime",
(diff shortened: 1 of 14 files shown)


src/pages/app/policy/list.tsx
```diff
@@ -102,7 +61,7 @@ export default function PoliciesList() {
         </Dialog>
       </PageHeader>
 
-      {policiesLoading || productsLoading ? (
+      {policiesLoading || columns.length === 0 ? (
```
ezekg: Not sure about this change, since afaict `columns` will never be empty.

What if the hook returned a tuple like other hooks, e.g.:

```ts
const [columns, {isLoading}] = usePolicyTableColumns()
```

Could leave it up to each hook to decide what to return after the columns, or use a common pattern.

src/hooks/use-policy-table-columns.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { useListProducts } from "@/queries/products"
+
+import { createTableColumnHelper } from "@/lib/tables"
+
+import ClipboardButton from "@/components/clipboard-button"
+
+const column = createTableColumnHelper<Policy>()
+
+export function usePolicyTableColumns() {
+  const { data: products = [] } = useListProducts()
```
ezekg: We should eventually refactor this to fetch each product individually, as some accounts have a large number of products that would need to be paginated, and iterating all of them to find the policy's is inefficient.

ezekg (changes requested):
I like this pattern a lot. Really cleans things up. Couple comments before merging.

src/hooks/use-policy-table-columns.tsx
Cazden: Makes sense! Added resource-specific cell components that handle their own fetching in https://github.com/keygen-sh/keygen-portal/pull/28/commits/884ece10415f81523409d1da70cac833b333f872, which queries by id instead of fetching all.

Also addresses your other https://github.com/keygen-sh/keygen-portal/pull/28#discussion_r2581594260 by moving loading logic to the cell itself.

src/queries/entitlements.ts
```diff
@@ -12,20 +17,22 @@ import { APIError } from "@/types/api"
 import * as keygen from "@/keygen"
 import { diff } from "@/lib/utils"
 
-export function useGetEntitlement(entitlementId: string) {
+export function useGetEntitlement(entitlementId: string | undefined) {
```
ezekg: In what cases would this (and others) be called with an undefined value? Seems like a code smell imo. Should probably be handled at the call site instead of allowing/handling the undefined value here.

ezekg (dismissed):
Looks good. One question, but feel free to merge!

src/queries/entitlements.ts
Cazden: I honestly wasn't sure if this should be handled here or upstream, but figured it would make call sites cleaner if it was in the query. Reverted this change https://github.com/keygen-sh/keygen-portal/pull/28/commits/4aa5f892c339389876ae455037cb828e543be6a3, it's handled at the call site now.

ezekg (approved):
This is such a good refactor! So clean. LGTM.

PR #41, opened 2025-12-06
Refactor older forms with field headers

### LINKED ISSUE #37: Update older forms to use standardized field header component
Forms prior to the introduction of the `<Field.Header />` component need their fields refactored to use it.

Cazden:
Standardizes older fields in `products` and `environments` forms with `<Field.Header />` components as used in newer forms. Also refactors/standardizes form field descriptions for the aforementioned resources as well as `entitlements`.

This will also close #37.


ezekg (approved):
Looks good, now stop working so we can play ARC Raiders.

PR #29, opened 2025-12-02
Refactor form schemas and fix linting errors

Cazden:
Fixes linting errors across codebase, refactors form schemas with the `BaseSchema` pattern that was used in policies, along with a few component refactors for table skeletons and doc links.

There's a few more housekeeping tasks I want to do but I'm going to separate them into another PR, this one's getting a bit big.

PR #44, opened 2025-12-10
Add policies api integration

Cazden:
Adds CRUD for policies and replaces mock policies data with API queries. Also fixes attribute values rendering with quotes/JSON, which was a regression caused by https://github.com/keygen-sh/keygen-portal/pull/29/commits/966027917ab29bb386329b071839078a2fc33e61.


src/keygen/policies/create.ts
```diff
@@ -0,0 +1,118 @@
...
+      relationships: {
+        product: {
+          data: {
+            type: "product",
+            id: productId,
+          },
+        },
+      },
+    },
+  }
```
Cazden: Going to refactor this so it's not so verbose. PR will be ready for review after.

ezekg: Maybe a `compact` util would be helpful here?

```ts
// remove null/undefined from arrays and plain objects
export type Compact<T> =
  T extends (infer U)[]
    ? NonNullable<U>[]
    : T extends object
      ? {
          [K in keyof T as T[K] extends null | undefined ? never : K]:
            NonNullable<T[K]>
        }
      : T

export function compact<T>(
  value: (T | null | undefined)[]
): NonNullable<T>[]
export function compact<T extends object>(
  value: T
): Compact<T>
export function compact(value: unknown): unknown {
  if (Array.isArray(value)) {
    return value.filter(
      (item): item is NonNullable<typeof item> => item != null
    )
  }

  if (value && typeof value === "object") {
    const result: Record<string, unknown> = {}

    for (const [key, val] of Object.entries(value)) {
      if (val != null) {
        result[key] = val
      }
    }

    return result
  }

  throw new TypeError("compact() expects an array or object")
}
```

Written by AI. Could probably be cleaned up, but hopefully gets the idea across.

```ts
compact({
  duration,
  strict,
  floating,
  scheme,
  requireProductScope,
  requirePolicyScope,
  requireMachineScope,
  requireFingerprintScope,
  requireComponentsScope,
  requireUserScope,
  ...
})
```

ezekg: Maybe even something like this could be cleaner and require less maintenance? Less explicit, but since it's all strongly typed, maybe that doesn't actually matter too much.

```ts
export default async function create({
  name,
  product: { id: productId },
  ...attributes
}: Forms.Policies.CreatePayload): Promise<PolicyResponse> {
  const body = {
    data: {
      type: "policies",
      attributes: {
        name,
        ...compact(attributes),
      },
      relationships: {
        product: {
          data: {
            type: "product",
            id: productId,
          },
        },
      },
    },
  }

  const result = (await client.request(`/accounts/${config.id}/policies`, {
    method: "POST",
    body: JSON.stringify(body),
  })) as PolicyResponse

  return result
}
```

Might end up biting us though, e.g. if we add a new relationship, idk?

Cazden: Added this and standardized it across the other resources too.

src/lib/compact.ts
```diff
@@ -0,0 +1,29 @@
...
+    return (value as unknown[]).filter(
+      (item): item is NonNullable<typeof item> => item != null,
+    )
+  }
+
+  if (value && typeof value === "object") {
+    const result: Record<string, unknown> = {}
+
+    for (const [key, val] of Object.entries(value)) {
+      if (val != null && val !== 0) {
```
ezekg: Typically, a compact function only removes null values. I'd filter out zero values another way instead. But tbh, it feels like a bug that the attribute is even being set to 0 by default in the first place without user input...? Mainly say that because 0 is not a valid value for heartbeat duration, since it should be at least 60 i.e. a minute.

At the very least, I'd default to 10 minutes if it must have a default numeric value, or simply default to null and let the API handle applying the default of 10 minutes when null.

ezekg (changes requested):
In addition to the compact/zero heartbeat duration issue (which, as noted, seems like a bug that maybe needs to be handled in the schema and not in the compact util), here are videos of the other issues I'd like to see tackled since they're directly related to functionality of policies:

### Heartbeat duration field does not accept input

https://github.com/user-attachments/assets/ed9ba966-fd8c-45fa-93b6-30d22ee6b6a7

### Policy duration cannot be edited

https://github.com/user-attachments/assets/bcef9475-2f40-4adc-8619-99372683dd44

### Policy edit form flickers when entitlements are changed

https://github.com/user-attachments/assets/7b2f55a8-20e5-43e9-9b31-c5465a48228a

Lastly, using `compact` on all attributes for the edit form may not actually be the best option, since a user may e.g. want to update a Timed policy with a duration to be a Perpetual policy without a duration — and that would be accomplished by setting the duration to null. Other update attributes are nullable as well, like max machines/users/etc.

Any attribute that has `allow_nil` [here](https://github.com/keygen-sh/keygen-api/blob/master/app/controllers/api/v1/policies_controller.rb#L125-L171) are nullable on update, i.e. setting to null "unsets" the value. Not sure how you want to accomplish that, but could be something worth adding to the update schema?

src/lib/utils.ts
```diff
@@ -8,6 +8,8 @@ export function cn(...inputs: ClassValue[]) {
 export function deepEqual(a: unknown, b: unknown): boolean {
   if (a === b) return true
 
+  if (!a && !b) return true
```
ezekg: This doesn't seem right to me. If this util is only meant for asserting the equality of objects and arrays, then it should never accept a number or other scalar types in the first place. If it is supposed check equality of scalars, then this guard clause is incorrect, because then 0 and false, or false and null, are equal.

src/lib/compact.ts
ezekg: fwiw, I think it's a bug with the heartbeat duration input field, not the schema. I see the default of 60 in the schema, but even when the field is set, it always shows as 1 second in the edit and duplicate forms.

image

As you can see, the response has `heartbeatDuration` set to 60, but the edit form shows 1 second, which is invalid.

ezekg: Seems to be because this nulls the field, which seems wrong. I don't think we should ever null fields that have values. The UI should always represent what the API sends, otherwise we'll hit state desync issues like this.

https://github.com/keygen-sh/keygen-portal/blob/6ef0240c78e8206cdf87157bc20ebc93979ef40e/src/components/policies/fields/all.tsx#L188-L201

Not really sure what all these `useEffects` are doing tbh. Maybe only relevant on create?

src/components/policies/fields/all.tsx
```diff
@@ -56,14 +56,67 @@ import KeyValueInput from "@/components/key-value-input"
 import NullableSelect from "@/components/nullable-select"
 import DurationInput, { HeartbeatPresets } from "@/components/duration-input"
 
+export type PolicyFieldName =
```
ezekg: This seems like it's going to be a chore to maintain as the policy schema evolves. I'd recommend:

```ts
export type PolicyFieldName =
  | keyof Forms.Policies.BaseValues
  | keyof Forms.Policies.CreateValues
  | keyof Forms.Policies.UpdateValues
```

Or instead of the union here and elsewhere, define `AllValues`?

src/lib/compact.ts
ezekg: tbh this feels like an incorrect use of `useEffect`: https://react.dev/learn/you-might-not-need-an-effect

Feel like something like this may be more obvious?

```tsx
<FormField
  control={form.control}
  name="requireHeartbeat"
  render={({ field }) => (
    <Select
      value={...}
      onValueChange={(value) => {
        field.onChange(value)

        // clear dependent fields directly in the event handler
        if (mode === Mode.Create && value == null) {
          form.resetField("heartbeatDuration")
          form.resetField("heartbeatCullStrategy")
          form.resetField("heartbeatBasis")
          form.resetField("heartbeatResurrectionStrategy")
        }
      }}
    />
  )}
/>
```

src/components/policies/fields/all.tsx
Cazden: Agreed! Refactored to `AllValues` and moved it to `/forms/policies` in https://github.com/keygen-sh/keygen-portal/pull/44/commits/7bb2e9f2cd850a705a5158dee3ecdde21df488e7.

src/lib/utils.ts
Cazden: Yeah I was unsure about this and https://github.com/keygen-sh/keygen-portal/pull/44#discussion_r2612781205 but it was working when I was doing some manual form testing at the time, should've spent more time on it. Fixed it in https://github.com/keygen-sh/keygen-portal/pull/44/commits/496cbef8520d27e34cba5f5393412fa2ae109152 and made some changes to `duration-input` to handle `null` values so they don't get transformed in https://github.com/keygen-sh/keygen-portal/pull/44/commits/97b945bc14bd2137f7ace5376f1050fe10d50f2e.

This also resolves your https://github.com/keygen-sh/keygen-portal/pull/44#discussion_r2614784638 too.

src/lib/compact.ts
Cazden: Really appreciate the detailed feedback and videos, helped a lot. All of this has been resolved in the latest commits, feel free to reopen this if you think any of this wasn't fixed though.

So it doesn't get buried, I want to point out that I solved sending `null` values to the API by just removing the use of `compact` for the update flow, because we already `diff` in the query which preserves `null` values properly. Changes in https://github.com/keygen-sh/keygen-portal/pull/44/commits/7e10d6cbfada1f3b335aca406d920b9c834a11ec.

src/forms/policies.ts
```diff
@@ -624,22 +638,26 @@ export function composePolicySchema(selection: {
...
+export function getSchemaDefaults<T extends BaseValues>(
+  schema: z.ZodType<T>,
+): T {
   const parsed = schema.parse({ name: "temp", product: { id: "temp" } })
-
   parsed.name = ""
-  parsed.product.id = ""
-
+  if ("product" in parsed) {
+    ;(parsed as CreateValues).product.id = ""
```
ezekg: Why is this so complex? You know `product` exists, because you define it a few lines above.

src/forms/policies.ts
```diff
@@ -694,4 +712,13 @@ export function getFormValuesFromPolicy(policy: Policy): BaseValues {
...
   }
+
+  if (options?.product) {
+    return {
+      ...base,
+      product: { id: policy.relationships.product?.data?.id ?? "" },
+    } as unknown as T
+  }
+
+  return base as unknown as T
```
ezekg: Seems like a code smell to cast to unknown when you already know its type

src/forms/policies.ts
ezekg: Maybe simplify to this?

```ts
export function getSchemaDefaults<T extends BaseValues>(
  schema: z.ZodType<T>,
): T {
  const parsed = schema.parse({ name: "temp", product: { id: "temp" } }) as AllValues
  parsed.name = ""
  parsed.product.id = ""
  return parsed as T
}
```

Also might be worth adding a comment on why those 2 default values are there at all. I remember there being a reason, but I can't remember the exact reason. Would be helpful to document.

src/forms/policies.ts
ezekg: Maybe do this so intent is clearer?

```ts
// ...

if (options?.product) {
  return {
    ...base,
    product: { id: policy.relationships.product?.data?.id ?? "" },
  } as AllValues as T
}

return base as T
```

src/components/duration-input.tsx
ezekg, line 84: When going from Unlimited to a preset, e.g. Days, can we default the number to 1? Right now you have to input a number before changing from Unlimited, otherwise it just reverts back to Unlimited.

https://github.com/user-attachments/assets/0c7692d4-edb0-4ae3-b06d-20a52e2d21da

ezekg: I tried this and it worked, but lmk if I'm missing any edge cases:

```ts
const onUnitChange = (u: Unit) => {
  setUnit(u)
  setUnitsOpen(false)

  // default when switching from unlimited to another unit
  if (num == null && u.seconds != null) {
    setNum(1)
    apply(1, u)
  } else {
    apply(num, u)
  }
}
```

src/components/duration-input.tsx
```diff
@@ -169,27 +172,38 @@ export default function DurationInput({
     setUnit(selectedUnit)
 
     const nextNum =
-      selectedUnit.seconds == null
-        ? 1
-        : typeof value === "number" && value > 0
+      value == null || selectedUnit.seconds == null
+        ? null
+        : value > 0
           ? Math.max(1, Math.round(value / selectedUnit.seconds))
```
ezekg: This is all very complex. I'd prefer to avoid nested ternary conditions in the future.

```ts
useEffect(() => {
  // recalculate num based on current unit, but preserve the unit
  if (value != null && value > 0 && unit.seconds != null) {
    setNum(Math.max(1, Math.round(value / unit.seconds)))
  } else {
    setNum(null)
  }
}, [value, unit])
```

This also fixes a bug where the UI resets the unit to Unlimited when backspacing the value:

https://github.com/user-attachments/assets/6b110ca4-aaf0-426f-93aa-2501f7f6f89e

We should preserve the unit unless explicitly changed. If the user wants unlimited, they can select it.

ezekg: Here's what it looks like with the above fixes:

https://github.com/user-attachments/assets/75841ff4-5dca-496f-be1d-0f5bf3aa0e4d

src/pages/app/policy/details.tsx
```diff
@@ -295,7 +282,7 @@ export default function PolicyDetails() {
                     Lease-based
                   </Badge>
                 )}
-                {isFeatureBased(policy) && (
+                {isFeatureBased(entitlements.length) && (
```
ezekg: Nit: I would pass in the entitlements array instead of an integer. That way in the future we can e.g. check if entitlements are active, or do other things. Just seems weird to pass in an integer here. As a general rule of thumb, it's better to pass around communicative values rather than uncommunicative values like raw integers.

Maybe I can eventually add a count to the policy's entitlements relationship, that way we can just pass in the policy and check `policy.relationships.entitlements.meta.count > 0`.

ezekg (changes requested):
Looks good. Couple bugs need to be fixed before merging.

src/forms/policies.ts
Cazden: Our schema requires `name` to have a min value of 1, and `product.id` is also required, so this is basically filling both values before we parse so Zod can parse without crashing, then we empty both values so the form isn't initialized with "temp" in both fields.

The conditional is there because `product.id` no longer exists on `BaseValues` after I separated it. However this function is only used in the create flow anyway, so I renamed the function to `getCreateSchemaDefaults` and updated it to use `CreateValues` instead of base so we can make this easier to read. Fixed in https://github.com/keygen-sh/keygen-portal/pull/44/commits/1216da13786e1f067eae7ae8355d5357d0c078b7.

src/pages/app/policy/details.tsx
Cazden: Yeah something like that would be neat! Refactored in https://github.com/keygen-sh/keygen-portal/pull/44/commits/3162b735b582161cbcaca08ec401332462e457a0.

src/components/duration-input.tsx
Cazden: Nice, looks good. I pushed your suggested fix in https://github.com/keygen-sh/keygen-portal/pull/44/commits/ec3157a7dda102cdb01aab4a129930c6ad6019bb. One thing I wanted to mention is the form behavior on a `Timed` policy template showed an error message if the `duration` field was empty that read `Cannot be unlimited for Timed policies`. This looked a little awkward imo when the unit was something like `days`, i.e.

image

I switched the error message to `Cannot be empty for Timed policies` instead as a simple fix, but if you want to slightly alter the behavior a bit to where we switch to unlimited on blur, that's also an option. Kind of violates what you mentioned about preserving the unit unless explicitly changed though, so I just changed the message instead but lmk.

src/components/duration-input.tsx
Cazden: Yeah that LGTM! I pushed the fix https://github.com/keygen-sh/keygen-portal/pull/44/commits/50c4b1eb27a2db70189dceac078d116ee10da259.

ezekg (approved):
Looks good. LGTM! 👍

PR #46, opened 2026-01-06
Add licenses interface for CRUD actions

Cazden:
Adds base licenses interface for list and details pages, CRUD dialogs/forms, and mobile styles. Also added a Metadata component to reuse across resource pages, and refactored existing resources with it.


src/pages/app/license/list.tsx
```diff
@@ -0,0 +1,76 @@
...
+      {licensesLoading ? (
+        <Skeletons.Table />
+      ) : (
+        <DataTable<License>
+          data={MockLicenses}
+          columns={columns}
+          hideOnMobile={[
+            "id",
+            "attributes.key",
+            "relationships.product",
```
ezekg, line 67: ~~Let's include the policy here as well, before product.~~

src/pages/app/license/details.tsx
```diff
@@ -0,0 +1,758 @@
...
+                    </Property.Section>
+
+                    <Separator />
+
+                    <Property.Section title="Usage limits" className="p-4">
+                      <Property.Field
+                        icon={Cpu}
+                        label="cores"
+                        variant="reverse"
+                        value={getCoresLimitDisplay(license, policy, 0)}
```
ezekg, line 764: Can we add an indicator when a license overrides its policy's limits? Can simply check if `license.attributes.maxY !== policy.attributes.maxY` and if so, show an "overridden" badge or something.

src/components/licenses/fields/limits.tsx
```diff
@@ -0,0 +1,381 @@
...
+                    label="Max Machines"
+                    optional
+                    tooltip={LicenseFormFieldDescriptions.maxMachines}
+                  >
+                    <FormControl>
+                      <Input
+                        {...field}
+                        type="number"
+                        min={1}
+                        placeholder="Inherit from policy"
```
ezekg: Wonder if we could add the inherited value here for reference instead of this text?

src/components/licenses/fields/expiration.tsx
```diff
@@ -0,0 +1,228 @@
...
+                    </FormControl>
+                  </Field.Header>
+                  <FormMessage />
+                </FormItem>
+              )
+            }}
+          />
+
+          <p className="text-sm text-content-muted">
+            Leave empty to inherit duration from the policy. Select a date to
```
ezekg:
```suggestion
            Leave empty to set according to the policy's duration ({durationFromBasisInWords}). Select a date to
```

Where `durationFromBasisInWords` is e.g. "3 months from now" or "3 months from first activation" depending on the policy's expiration basis. Always good to remind the user what they're doing.

src/forms/licenses.ts
```diff
@@ -0,0 +1,59 @@
...
+    .trim()
+    .optional()
+    .refine(
+      (value) => !value || value.length >= 8,
+      "Key must be at least 8 characters",
+    ),
+  expiry: z.string().nullable().optional(),
+  policyId: z.string().min(1, "Policy is required"),
+  suspended: z.boolean().optional().default(false),
+  protected: z.boolean().optional().default(false),
```
ezekg: For a license, this should default to `null`, since it's inherited from the policy unless overridden.

src/pages/app/license/details.tsx
```diff
@@ -412,7 +412,7 @@ export default function LicenseDetails() {
                       value={
                         productError ? (
                           <Badge variant="destructive">ERROR</Badge>
-                        ) : productLoading || productFetching ? (
+                        ) : policyLoading || policyFetching || productLoading || productFetching ? (
```
ezekg: Why do we need to check the policy before displaying the product?

src/components/licenses/fields/general.tsx
```diff
@@ -0,0 +1,201 @@
...
+                          placeholder="Leave blank for auto-generation"
+                          className="font-mono"
+                        />
+                      </FormControl>
+                      <Button
+                        type="button"
+                        variant="outline"
+                        onClick={handleGenerateKey}
+                      >
+                        Generate
```
ezekg: We can drop the auto-generate key functionality. It's a legacy artifact and I don't want to support it in the new UI.

src/components/licenses/fields/general.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    >
+                      <FormControl>
+                        <SelectTrigger className="w-full">
+                          <SelectValue placeholder="Select a policy..." />
+                        </SelectTrigger>
+                      </FormControl>
+                      <SelectContent>
+                        {policies.map((policy) => (
+                          <SelectItem key={policy.id} value={policy.id}>
+                            {policy.attributes.name}
```
ezekg: Adding product name here too might be helpful to differentiate.

What if we did a grouped select by product? e.g.:

```
Product X
  Policy A
  Policy B
Product Y
  Policy C
```

src/pages/app/license/details.tsx
```diff
@@ -0,0 +1,758 @@
...
+              <div className="mt-2 flex flex-col gap-1 text-sm text-content-subdued md:flex-row md:items-center md:gap-4">
+                <Button
+                  variant="clipboard"
+                  size="clipboard"
+                  onClick={() => copyToClipboard(license.attributes.key)}
+                  className="w-fit border-none"
+                >
+                  <Key className="mr-1 size-4" />
+                  <span className="w-64 truncate font-mono md:w-auto">
+                    {license.attributes.key}
```
ezekg: Not sure I like the key attribute here. But it might grow on me, so we can leave it for now. One thing I'd like to do is make this section better match the policy's product, which uses a similar pattern:

```tsx
<div className="mt-2 flex h-4 items-center text-sm text-content-subdued">
  <Key className="mr-2 size-4 pt-0.5" />
  <span>Key:</span>
  <div className="flex items-center gap-1">
    <Button
      variant="link"
      size="link"
      className="peer ml-3 text-content-muted"
      onClick={() => copyToClipboard(license.attributes.key)}
    >
      {license.attributes.key}
    </Button>
    <Copy className="mt-0.5 ml-2 size-3.5 transition-all duration-200 peer-hover:translate-x-2 peer-hover:text-brand-primary" />
  </div>
</div>
```

ezekg: Also need to add a max-width since some license keys, e.g. signed keys, can be very long:

Image

ezekg: Should probably truncate the keys past a certain length, formatting similarly to Bitcoin address and transactions so that it's easier to verify:

```ts
export function truncateKey(
  key: string,
  { maxLength = 64 }: { maxLength: number },
): string {
  if (key.length <= maxLength) return key

  const head = key.slice(0, Math.ceil(maxLength / 2))
  const tail = key.slice(-Math.floor(maxLength / 2))

  return `${head}…${tail}`
}
```

```tsx
<div className="mt-2 flex h-4 items-center text-sm text-content-subdued">
  <Key className="mr-2 size-4 pt-0.5" />
  <span>Key:</span>
  <div className="flex items-center gap-1">
    <Button
      variant="link"
      size="link"
      className="peer ml-3 text-content-muted"
      onClick={() => copyToClipboard(license.attributes.key)}
    >
      {truncateKey(license.attributes.key, { maxLength: isMobile ? 32 : 64 })}
    </Button>
    <Copy className="mt-0.5 ml-2 size-3.5 transition-all duration-200 peer-hover:translate-x-2 peer-hover:text-brand-primary" />
  </div>
</div>
```

Image

src/pages/app/license/details.tsx
```diff
@@ -0,0 +1,758 @@
...
+                              navigate({
+                                to: "/$id/app/products/$productId",
+                                params: {
+                                  id: keygen.config.id,
+                                  productId: product.id,
+                                },
+                              })
+                            }
+                          >
+                            {product.attributes.name}
```
ezekg: These relationship buttons need a pointer cursor.

src/components/licenses/edit/edit-form.tsx
```diff
@@ -0,0 +1,91 @@
...
+            className="max-w-48 flex-1 basis-1/2"
+          >
+            Cancel
+          </Button>
+          <Button
+            type="submit"
+            disabled={loading}
+            className="max-w-48 flex-1 basis-1/2"
+          >
+            {loading ? <Loading.Dots className="bg-background" /> : "Update"}
```
ezekg, line 90: The edit form's submit button isn't working for me.

src/pages/app/license/details.tsx
```diff
@@ -0,0 +1,758 @@
...
+                    <Skeleton className="h-4 w-3/4" />
+                    <Skeleton className="h-4 w-1/2" />
+                    <Skeleton className="h-4 w-1/3" />
+                  </div>
+                )}
+              </TabsContent>
+
+              <TabsContent value="events" className="p-4"></TabsContent>
+            </SidebarContent>
+            <SidebarFooter className="p-4"></SidebarFooter>
```
ezekg: Missing the "View All Attributes" button.

ezekg (changes requested):
Looking good. Have a handful of notes.

src/pages/app/license/list.tsx
ezekg: Sorry, disregard. I misunderstood what this was doing. I thought it was defining columns, but now I see it's what's hidden on mobile. I think policy and product are important for finding the right license, and so shouldn't be hidden on mobile. Also, ID is the most important attribute for finding resources, so I don't think it should ever be hidden.

I think adding horizontal scroll is fine, and there's no way to avoid it for e.g. licenses with long names, long policy names, etc. so might as well assume it'll happen. I always thought [this scroll behavior](https://x.com/pontusab/status/1992162748134989999) with the arrows was rad.

src/hooks/use-license-table-columns.tsx
```diff
@@ -0,0 +1,75 @@
...
+        cell: (info) => <ClipboardButton value={info.getValue()} />,
+      }),
+      column.attr("name", {
+        header: "Name",
+        cell: (info) =>
+          info.getValue() || <span className="text-content-muted">--</span>,
+      }),
+      column.attr("key", {
+        header: "Key",
+        cell: (info) => <Badge className="font-mono">{info.getValue()}</Badge>,
```
ezekg: I'd prefer the key to match the ID's code-style badge for consistency. Right now it's inconsistent:

Image

Also need to be sure we handle long keys here as well, so need to truncate similarly to the details page. Not sure if it should apply to the clipboard button specifically, but you could add [a truncation style](https://gist.github.com/ezekg/b7308aa5dd9c94107d5985b6041b6301) for keys.

But this adds some width-related issues to the button, so will need to fix that. Looks cleaner, tho:

Image

ezekg: Fixed the width issue in the latest gist: https://gist.github.com/ezekg/b7308aa5dd9c94107d5985b6041b6301

ezekg: I don't like how the table rows change widths when these are hovered, but not sure how to fix. So I'll let you figure it out from here. 😅

src/pages/app/license/details.tsx
Cazden: I reversed this change in favor of https://github.com/keygen-sh/keygen-portal/pull/46/commits/b9b1064f75b96954055b9caaba72bcbc8ff96103 (prevents query with an empty id) because this was just verbose to me, but this was originally because the app was crashing on page load because the product id was dependent on the policy, so if policy hadn't finished fetching yet, product would've still returned although `undefined` because `productId` was an empty string, then this `productLoading` condition would've succeeded and crashed the app because `product.attributes` doesn't exist on `undefined`.

src/components/licenses/fields/general.tsx
Cazden: Definitely! I think it looks good:
image

src/hooks/use-license-table-columns.tsx
Cazden: Yeah this one was annoying, don't blame you for throwing in the towel lol. Implemented the changes in your gist along with fixing the non-fixed width that was causing the table resizing in https://github.com/keygen-sh/keygen-portal/pull/46/commits/6e6b0961a9ca4cc27c5d0f8c5cacb2c93cee59b7.

src/pages/app/license/details.tsx
Cazden: This was implemented in https://github.com/keygen-sh/keygen-portal/pull/46/commits/6e6b0961a9ca4cc27c5d0f8c5cacb2c93cee59b7 along with some minor changes in the following commits.

ezekg (approved):
Looks good. Once we wire up the API, I'd like for the license attributes to dynamically display relevant information based on the policy. E.g. if the license uses a node-locked, then display max X attributes in it's own box (because the overview info in the sidebar is easy to miss), if it's a timed license, display expiry and expiration basis/etc.

But we can do that stuff later. This is a good base to build upon.

PR #48, opened 2026-01-12
Add groups interface for CRUD actions

Cazden:
Adds base interface for the Groups resource that includes list and details pages, CRUD dialogs/forms, and mobile styles.


ezekg (approved):
Looks good. 👍

PR #49, opened 2026-01-14
Add machines interface for CRUD actions

Cazden:
Adds base interface for the Machines resource that includes list and details pages, CRUD dialogs/forms, and mobile styles. Also adds a new Search Select component for searching resources by ID/name or custom parameters per resource (e.g. licenses would be searchable by id or name, or key if name isn't set).

### DIFF
--- a/src/components/search-select.tsx
+++ b/src/components/search-select.tsx
@@ -1,0 +1,260 @@
+import { useState, useRef, useMemo } from "react"
+
+import { Input } from "@/components/ui/input"
+import { Button } from "@/components/ui/button"
+import { ScrollArea } from "@/components/ui/scroll-area"
+import { Command, CommandList, CommandItem } from "@/components/ui/command"
+import {
+  Popover,
+  PopoverTrigger,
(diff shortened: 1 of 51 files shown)


ezekg (approved):
Looks great! LGTM!

src/components/search-select.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+interface ResourceConfig {
+  getLabel: (option: BaseOption) => string
+  placeholder: string
+  searchPlaceholder: string
+  emptyMessage: React.ReactElement
+}
+
+const resourceConfigs: Record<ResourceType, ResourceConfig> = {
+  license: {
```
ezekg, line 37: This probably won't be maintainable as we add resources. Making it a composable higher-order component with e.g. a label, placeholders, data source, etc. pieces will make this easier to maintain as the resource list grows. Then you can have `license-select`, etc. as importable components that use this higher-order search component. But I think we already talked about that iirc, so fine to handle in a separate PR.

PR #52, opened 2026-01-20
Add users interface for CRUD actions

Cazden:
Adds base interface for the Users resource that includes list and details pages, CRUD dialogs/forms, and mobile styles. Also adds a couple new components—ConfirmationModal and TooltipSelectItem.
* ConfirmationModal replaces DeleteModal as a more generic component that we can use for other alert modals, like in the case of Users, the "Ban" or "Reset Password" actions:
image

* And TooltipSelectItem adds hoverable tooltips to select input items:
image


ezekg (approved):
Looks good. 👍

PR #53, opened 2026-01-24
Add Licenses API integration for CRUD actions

Cazden:
Adds API integration for License CRUD actions and removes mock License data across all resources. Also extends the Input component with a disabled tooltip option, which is used in license forms for letting users know why the `maxMachines` form field is disabled when trying to create a license using a non-floating policy.

### DIFF
--- a/src/queries/licenses.ts
+++ b/src/queries/licenses.ts
@@ -1,0 +1,117 @@
+import { useQuery, useQueryClient, useMutation } from "@tanstack/react-query"
+
+import { useEnvironment } from "@/hooks/use-environment"
+
+import * as Forms from "@/forms"
+
+import { APIError } from "@/types/api"
+import { License } from "@/types/licenses"
+
+import * as keygen from "@/keygen"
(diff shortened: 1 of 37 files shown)


src/queries/licenses.ts
```diff
@@ -0,0 +1,116 @@
...
+    queryFn: async () => {
+      const response = await keygen.licenses.get({ id: licenseId })
+
+      if (!response.data) {
+        throw new Error("License not found")
+      }
+
+      return response.data
+    },
+    retry: (failures, error) =>
```
ezekg: Let's not implement retry logic yet. Too easy to introduce [an API DoS vector](https://medium.com/@jonathanboice/cloudflare-accidentally-ddosd-itself-with-react-lessons-for-frontend-engineers-2538dcd757b1).

ezekg (changes requested):
One note on retry logic. Other than that, g2g.

ezekg (approved):
g2g

PR #55, opened 2026-01-26
Add Groups API integration

Cazden:
Adds API integration for Groups and removes mock group data across all resources.


ezekg (approved):
g2g

PR #56, opened 2026-01-27
Add standardized form API error handling

Cazden:
Need a way to standardize form error handling across resources, so e.g. in a multistep flow like creating a machine, if the API rejects with a specific error (i.e. `MACHINE_LIMIT_EXCEEDED`), we automatically navigate back to the form step that includes the field related to the error to display a message to the user that's helpful, like "Machine limit exceeded for this license" on the `License Relationship` field. This PR addresses this issue so we have a reusable/standardized format for: selecting what fields to invalidate, what message to display, and what error code is required to trigger it.

Example usage:

```
...
const steps: Step[] = useMemo(
    () => [
      {
        key: Steps.General,
        title: "General configuration",
        fields: ["fingerprint", "licenseId"],
        render: () => <Machines.Fields.General />,
      },
      ...
    ],
    [],
)
...

const handleCreateMachine = useCallback(
    (values: Forms.Machines.CreateValues) => {
      createMachine.mutate(values, {
        onSuccess: (machine) => {
          toast({ message: "Machine activated", variant: "success" })
          onSelectMachine(machine)
          onClose()
        },
        onError: (error) => {
          handleFormError({
            form,
            error,
            steps, // Where we access the fields, defined above 
            goToStep: goTo,
            toastMessage: "Failed to activate machine",
            fieldMappings: [
              {
                code: MachineErrorCode.MachineLimitExceeded, // MACHINE_LIMIT_EXCEEDED
                field: "licenseId", // From the fields in 'steps' so it's type-safe
                message: "Machine limit exceeded for this license",
              },
            ],
          })
        },
      })
    },[],
 )
  ```

### DIFF
--- a/src/lib/form-errors.ts
+++ b/src/lib/form-errors.ts
@@ -1,0 +1,130 @@
+import { FieldPath, FieldValues, UseFormReturn } from "react-hook-form"
+
+import { APIError } from "@/types/api"
+
+import { toast } from "@/lib/toast"
+
+export type ErrorFieldMapping<TFieldValues extends FieldValues> = {
+  code: string
+  field: FieldPath<TFieldValues>
+  message?: string
+}
+
+export type FormStep<TFieldValues extends FieldValues> = {
+  key: string
+  fields?: FieldPath<TFieldValues>[]
+}
+
+export type HandleFormErrorOptions<TFieldValues extends FieldValues> = {
+  form: UseFormReturn<TFieldValues>
+  error: APIError
+  fieldMappings?: ErrorFieldMapping<TFieldValues>[]
+  steps?: FormStep<TFieldValues>[]
+  goToStep?: (index: number) => void
+  toastMessage: string
+  showToast?: boolean
+}
+
+function extractFieldFromPointer(pointer: string | undefined): string | null {
+  if (!pointer) return null
+
+  const attributesMatch = pointer.match(/\/data\/attributes\/(\w+)/)
+  if (attributesMatch) return attributesMatch[1]
+
+  const relationshipsMatch = pointer.match(/\/data\/relationships\/(\w+)/)
+  if (relationshipsMatch) {
+    return `${relationshipsMatch[1]}Id`
+  }
+
+  return null
+}
+
+function findStepForField<TFieldValues extends FieldValues>(
+  field: FieldPath<TFieldValues>,
+  steps: FormStep<TFieldValues>[],
+): number {
+  const fieldString = String(field)
+
+  for (let i = 0; i < steps.length; i++) {
+    const step = steps[i]
+    if (!step.fields) continue
+
+    for (const stepField of step.fields) {
+      if (stepField === field) return i
+      if (fieldString.startsWith(String(stepField))) return i
+    }
+  }
+
+  return -1
+}
+
+export function handleFormError<TFieldValues extends FieldValues>(
+  options: HandleFormErrorOptions<TFieldValues>,
+): boolean {
+  const {
+    form,
+    error,
+    fieldMappings = [],
+    steps = [],
+    goToStep,
+    toastMessage,
+    showToast = true,
+  } = options
+
+  let fieldSet = false
+  let targetField: FieldPath<TFieldValues> | null = null
+  let errorMessage: string | null = null
+
+  if (error.code) {
+    const mapping = fieldMappings.find((mapping) => mapping.code === error.code)
+    if (mapping) {
+      targetField = mapping.field
+      errorMessage = mapping.message ?? error.detail ?? "Field is invalid"
+    }
+  }
+
+  if (!targetField && error.source?.pointer) {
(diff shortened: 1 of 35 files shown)

### CODE REVIEW

File: src/lib/form-errors.ts [hunk 1]
```diff
@@ -0,0 +1,130 @@
...
+      if (fieldString.startsWith(String(stepField))) return i
+    }
+  }
+
+  return -1
+}
+
+export function handleFormError<TFieldValues extends FieldValues>(
+  options: HandleFormErrorOptions<TFieldValues>,
+): boolean {
```
ezekg: Not sure it needs to return a boolean? Probably fine to return void here.

src/lib/form-errors.ts
```diff
@@ -0,0 +1,130 @@
+import { FieldPath, FieldValues, UseFormReturn } from "react-hook-form"
+
+import { APIError } from "@/types/api"
+
+import { toast } from "@/lib/toast"
+
+export type ErrorFieldMapping<TFieldValues extends FieldValues> = {
```
ezekg: I'm not sure you're going to be able to accurately map error codes to fields like this. The `MACHINE_LIMIT_EXCEEDED` error code, in particular, can be triggered not only when a license exceeds its machine limit, but also when [a user does](https://github.com/keygen-sh/keygen-api/blob/master/features/api/v1/machines/create.feature#L4188-L4195). This can occur when the policy's [machine leasing strategy](https://keygen.sh/docs/api/policies/#policies-object-attrs-machineLeasingStrategy) is `PER_USER`. Right now, the error message is the only place we communicate whether the limit was exceeded by the license or the user.

To be honest, it's probably fine to let errors without a pointer to a specific attribute or relationship, e.g. `/data`, be a "general" error that's displayed verbatim, and which that doesn't navigate steps.

Doing it that way should really simplify the error handling logic, which is a bonus.

Cazden:
General/verbatim error:
image

Pointer that navigates back to step/field:
image

Cazden:
With the changes in c865dd5 and fda1f57 that accommodates our forms refactor (#70), usage now makes form submit handlers much more lightweight, as error handling is internalized in the `Wizard` and `Sheet` form layouts.

e.g. Previous submit handler example:
```
const handleSubmit = useCallback(async () => {
  await form.handleSubmit((values) => {
    createProduct.mutate(values, {
      onSuccess: async (product) => {
        toast({ message: "Product created", variant: "success" })
        onOpenChange(false)
        await navigateToResource(product)
      },
      onError: (error) => {
        toast({
          message: "Failed to create product",
          description: error.detail,
          variant: "error",
        })
      },
    })
  })()
}, [form, createProduct, navigateToResource, onOpenChange])
```

New submit handler:
```
const handleSubmit = useCallback(async () => {
  const values = form.getValues()
  const product = await createProduct.mutateAsync(values)
  toast({ message: "Product created", variant: "success" })
  onOpenChange(false)
  await navigateToResource(product)
}, [form, createProduct, navigateToResource, onOpenChange])
```

Internally this is what's happening in the form layout components:
```
const valid = await form.trigger()

// Catch schema errors
if (!valid) {
  await handleFormError({
    form,
    steps: steps.map((step) => ({ key: step.id, fields: step.fields })),
    goToStep: goToIndex,
    toastMessage: errorMessage ?? "",
  })
  return
}

try {
  await onSubmit()
} catch (error) {
  // Catch API errors
  if (errorMessage && error instanceof APIError) {
    await handleFormError({
      form,
      steps: steps.map((step) => ({ key: step.id, fields: step.fields })),
      goToStep: goToIndex,
      toastMessage: errorMessage ?? "",
      apiError: error,
    })
  }
}
```

ezekg (dismissed):

src/lib/form-errors.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Iterates through all form steps for the step that contains the specified field
+function findStepForField<TFieldValues extends FieldValues>(
+  field: FieldPath<TFieldValues>,
+  steps: FormStep<TFieldValues>[],
+): number {
+  const fieldString = String(field)
+
+  for (let i = 0; i < steps.length; i++) {
```
ezekg, line 35: Nit: since you're not using `i` other than for `step`, this could be `for (const step of steps)` or `steps.forEach`.

src/lib/form-errors.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  return -1
+}
+
+// Sequentially triggers validation for all form steps until a step fails then returns its index
+async function findStepWithError<TFieldValues extends FieldValues>(
+  form: UseFormReturn<TFieldValues>,
+  steps: FormStep<TFieldValues>[],
+): Promise<number> {
+  for (let i = 0; i < steps.length; i++) {
```
ezekg, line 53: Nit: https://github.com/keygen-sh/keygen-portal/pull/56/changes#r2856711375

ezekg (approved):
Approved, with a couple nits that you're free to ignore.

PR #64, opened 2026-01-29
Add resource navigation hook

Cazden:
Adds a reusable and more maintainable way for navigating to specific resource pages by using a hook, and refactors existing navigate calls that go to a specific resource id.

Example:

```
const navigateToResource = useResourceNavigate()
...
return (
  <DataTable<Policy>
    ...
    onRowClick={(policy) => navigateToResource(policy, "policy")}
  />
)
 ```
 
Adding this with future refactoring in mind, where I'd like to reduce prop-drilling in forms. We're currently passing around a handler from the list view down to create modal/forms that handle navigating to the resource upon creation, so refactoring that functionality into a reusable hook would eliminate that prop and clean it up a bit.

### DIFF
--- a/src/hooks/use-resource-navigate.ts
+++ b/src/hooks/use-resource-navigate.ts
@@ -1,38 +1,20 @@
 import { useNavigate } from "@tanstack/react-router"
 
-import { AnyResource, ResourceType } from "@/types/api"
+import { AnyResource } from "@/types/api"
 
 import * as keygen from "@/keygen"
-
-const ResourcePlurals: Record<ResourceType, string> = {
-  component: "components",
-  entitlement: "entitlements",
-  group: "groups",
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: src/hooks/use-resource-navigate.ts [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+  product: "products",
+  user: "users",
+}
+
+export function useResourceNavigate() {
+  const navigate = useNavigate()
+
+  return async (
+    resource: AnyResource | null,
+    resourceType: ResourceType,
```
ezekg: Could probably use `resource.type` instead, no? That's already plural so would simplify this a lot by getting rid of the `resourceType` param.

ezekg (approved):

PR #68, opened 2026-02-02
Add Users API integration for CRUD action

Cazden:
Adds API integration for Users CRUD actions and removes mock users data across all resources.

### DIFF
--- a/src/keygen/users/list.ts
+++ b/src/keygen/users/list.ts
@@ -1,0 +1,39 @@
+import config from "@/keygen/config"
+import client from "@/keygen/client"
+import { UserListResponse, AllRoles } from "@/types/users"
+
+config.validate()
+
+interface ListProps {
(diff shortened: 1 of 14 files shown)


ezekg (approved):

src/keygen/users/list.ts
```diff
@@ -0,0 +1,39 @@
...
+    params.set("limit", limit.toString())
+  }
+  if (pageNumber != null) {
+    params.set("page[number]", pageNumber.toString())
+  }
+  if (pageSize != null) {
+    params.set("page[size]", pageSize.toString())
+  }
+
+  AllRoles.forEach((role) => params.append("roles[]", role))
```
ezekg, line 29: Would eventually like to separate admin vs non-admin users: https://github.com/keygen-sh/keygen-portal/issues/69

PR #70, opened 2026-02-14
Add forms component library to standardize forms

### LINKED ISSUE #59: Add standardized form/attribute descriptions
Attribute/form field descriptions for all resources should be defined in a single file per resource for easier maintenance. Currently, descriptions don't account for context differences (e.g. an edit form field might need a different description than the create form).

`LicenseAttributeDescriptions` used in the attributes page or as general attribute descriptions.
`LicenseFormFieldDescriptions` that inherits `LicenseAttributesDescriptions` for general form field overrides.
`LicenseCreateFormFieldDescriptions` that inherits `LicenseFormFieldDescriptions` for create form field overrides
`LicenseEditFormFieldDescriptions` that inherits `LicenseFormFieldDescriptions` for edit form field overrides

Wiring these up in their respective resources would standardize an easily maintainable source of attribute and form field descriptions, e.g. In the edit user flow (which already implements this pattern), we have `UserEditFormFieldDescriptions`, which has an override for the `password` attribute but no override for e.g. email:

```
export const UserEditFormFieldDescriptions: typeof UserFormFieldDescriptions = {
  ...UserFormFieldDescriptions,
  password: "Manually change the user's password. Leave the password blank if you do not wish to change it.",
}
```

So email would default to `UserFormFieldDescriptions` email key, which itself falls back to `UserAttributeDescriptions` of:
```
export const UserAttributeDescriptions: Readonly<
  Record<keyof Omit<UserAttributes, "created" | "updated">, string>
> = {
...
  email: "The unique email of the user.",
...
}
```

Which in the end gives us a clean override of `password` for the edit form, while preserving the description for `email`.

Cazden:
Addresses the mixed bag of components/patterns across forms from early on in the project vs. now, where our component library has matured, i.e. environments and products vs. policies and licenses, the latter utilizing newer components and better design choices. This PR standardizes all forms so they're easily understood and more composable, should any changes need to be made in the future, and makes styles/design/UX cohesive across all forms.

The core is a new form component library, which creates composable building blocks for layouts, sections, and fields. This includes a Wizard for multi-step flows with breadcrumbs/validation, a Sheet for simpler forms (e.g. edit forms which are usually single-step), and section primitives like Card, Columns, and Toggled (e.g. "advanced" toggle sections in the create policy flow).

Every resource should now follow the same `form/` structure: `create.tsx`, `edit.tsx`, and `fields.tsx`. The `fields.tsx` is a single source of truth for all field definitions with include/exclude props, replacing the old pattern of fields scattered across grouped files or inline in the forms themselves. This allows for dynamic grouping at the form level rather than pre-defined groupings, so it's a lot more flexible if we want to adjust what fields are included/excluded per form, without having to redefine the fields inline in every form. Validation/field dependency is still supported, i.e. heartbeat-related fields will be disabled if `requireHeartbeat` is false.

Some other changes:
* `src/forms/` renamed `to src/schemas/` to reflect what they actually are and free up the `forms` namespace for the new component library previously mentioned
* New `{Resource}FormFieldDescriptions` that closes #59
* Auto focus support added to some field components for wizard step transitions (i.e. `TagInput`, `MultiSelect`, etc.)


src/components/environments/form/fields.tsx
```diff
@@ -0,0 +1,216 @@
...
+  EnvironmentCreateFormFieldDescriptions,
+  EnvironmentEditFormFieldDescriptions,
+} from "@/types/environments"
+
+import * as Forms from "@/components/forms"
+import { CardSelector, CardOption } from "@/components/card-selector"
+
+type EnvironmentFieldName = "name" | "code" | "isolationStrategy"
+
+type FieldVariant = "row" | "stacking" | "inline" | "none"
```
ezekg: These probably don't need to be redefined for every file, unless you're planning on each resource being configured differently (doubtful). Probably best to put this into a general `src/components/form` directory or something so that it can be reused across forms.

src/components/environments/form/fields.tsx
```diff
@@ -0,0 +1,216 @@
...
+  schema,
+}: EnvironmentsFormFieldsProps) {
+  const descriptions =
+    schema === "create"
+      ? EnvironmentCreateFormFieldDescriptions
+      : schema === "edit"
+        ? EnvironmentEditFormFieldDescriptions
+        : EnvironmentFormFieldDescriptions
+
+  const fieldMap: Record<EnvironmentFieldName, React.ReactNode> = {
```
ezekg: This is inefficient since it instantiates every field, even if every field isn't used. This would likely be bad performance-wise for larger resources, like policies. I'd only instantiate the necessary fields.

You could probably use an exhaustive switch statement for this, tbh:

```tsx
export default function PoliciesFormFields({
  include,
  exclude = [],
  autoFocus,
  titleVariant,
  fieldVariant = "row",
  mode = PolicyMode.Create,
  schema,
}: PoliciesFormFieldsProps) {
  const form = useFormContext<Schemas.Policies.AllValues>()

  ...

  const fields = include
    ? include
    : DefaultFieldSort.filter((field) => !exclude.includes(field))

  return (
    <>
      {fields.map(field => {
        switch (field) {
          case "authenticationStrategy":
            return (
              <AuthenticationStrategyField
                key="authenticationStrategy"
                autoFocus={autoFocus === "authenticationStrategy"}
                fieldVariant={fieldVariant}
                descriptions={descriptions}
              />
            )
          case "checkInInterval":
            return (
              <CheckInIntervalField
                key="checkInInterval"
                autoFocus={autoFocus === "checkInInterval"}
                fieldVariant={fieldVariant}
                descriptions={descriptions}
                mode={mode}
              />
            )

          ...

          case "transferStrategy":
            return (
              <TransferStrategyField
                key="transferStrategy"
                autoFocus={autoFocus === "transferStrategy"}
                fieldVariant={fieldVariant}
                descriptions={descriptions}
                disabled={!duration}
              />
            )
          case "usePool":
            return (
              <UsePoolField
                key="usePool"
                autoFocus={autoFocus === "usePool"}
                descriptions={descriptions}
              />
            )
          default:
            return null
        }
      })}
    </>
  )
}
```

Maybe this is even be the cause of the performance issues on the policy edit page?

src/components/environments/form/fields.tsx
```diff
@@ -0,0 +1,216 @@
...
+    ),
+    code: (
+      <CodeField
+        key="code"
+        autoFocus={autoFocus === "code"}
+        fieldVariant={fieldVariant}
+        descriptions={descriptions}
+      />
+    ),
+    isolationStrategy: <IsolationStrategyField key="isolationStrategy" />,
```
ezekg: Is it intentional that this component is missing the other attributes?

ezekg (changes requested):
I haven't reviewed the whole thing, but this is my first pass. Once these are resolved I'll re-review.

src/components/environments/form/fields.tsx
Cazden: Yeah, I didn't think a selectable card component would need an auto focus, `fieldVariant` isn't applicable to `CardSelector` components, and the form field description are wired up differently than other form fields via a CardOption array:

```
const strategyOptions: CardOption<IsolationStrategy>[] = [
  {
    value: IsolationStrategy.Isolated,
    label: "Isolated",
    icon: <GlobeLock className="size-6 text-content-subdued md:size-5" />,
    tooltip: IsolationStrategyDescriptions[IsolationStrategy.Isolated],
  },
  {
    value: IsolationStrategy.Shared,
    label: "Shared",
    icon: <Globe className="size-6 text-content-subdued md:size-5" />,
    tooltip: IsolationStrategyDescriptions[IsolationStrategy.Shared],
  },
]
```

This also occurs in a few other cases, i.e. checked form fields omit fieldVariant because the alternative variants don't look good with them:

```
suspended: (
  <SuspendedField
    key="suspended"
    autoFocus={autoFocus === "suspended"}
    descriptions={descriptions}
  />
),
```

ezekg: Okay sg, was just checking to make sure it wasn't an oversight. 👍

src/components/forms/section/toggled.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const hasFieldErrors = fields.some((f) => {
+    const parts = f.split(".")
+    let current: unknown = errors
+    for (const part of parts) {
+      if (current && typeof current === "object" && part in current) {
+        current = (current as Record<string, unknown>)[part]
+      } else {
+        current = undefined
+        break
```
ezekg: This function is really hard to follow. First, a short comment on what this is actually doing would go a long way for cases like this, because it's not immediately clear. Second, I think it can be simplified:

```ts
const { control, getFieldState } = useFormContext<FieldValues>()
const state = useFormState({ control, name: fields })

// keep the section open when any field is invalid so the user can see and fix it
const open = showAdvanced || fields.some(
  (name) => getFieldState(name, state).invalid,
)
```

src/components/policies/form/create.tsx
```diff
@@ -0,0 +1,1094 @@
...
+                "duration",
+                "expirationStrategy",
+                "expirationBasis",
+                "renewalBasis",
+                "transferStrategy",
+              ]}
+            >
+              <Forms.Section.Card title="Timing">
+                <Policies.Form.Fields schema="create" include={["duration"]} />
+                <Forms.Section.Toggled>
```
ezekg, line 649: This should show some "advanced" text next to the toggle like it used to. Right now it's just a disembodied switch with no indication as to what it's used for (from the user's POV).

src/components/components/form/fields.tsx
```diff
@@ -0,0 +1,281 @@
...
+interface ComponentsFormFieldsProps {
+  include?: Schemas.Components.FieldNames[]
+  exclude?: Schemas.Components.FieldNames[]
+  autoFocus?: Schemas.Components.FieldNames
+  titleVariant?: boolean
+  fieldVariant?: FieldVariant
+  schema?: "create" | "edit"
+}
+
+const DefaultFieldSort: Schemas.Components.FieldNames[] = [
```
ezekg: The name of this const sounds like it should be used for sorting, but it's actually used (e.g. line 61) for default included fields. I'd maybe rename this to `DefaultFieldNames` or something else more descriptive.

src/components/forms/container/dialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+        disableOverlay={disableOverlay}
+        className={cn(
+          fullscreen || isMobile
+            ? "min-h-screen min-w-screen rounded-none border-none"
+            : "md:max-w-fit",
+          className,
+        )}
+      >
+        {/* Eliminates console warnings; form layouts handle visual headers */}
+        <DialogHeader className="sr-only">
```
ezekg, line 45: Assuming `sr-only` means screen reader only? Comment might want to clarify if so.

src/components/duration-input.tsx
ezekg: Maybe move this to the schema or something so that the warnings are fixed.

ezekg (changes requested):
Overall, great refactor. The new form primitives are a great DX improvement. Had a few notes.

src/components/components/form/fields.tsx
Cazden: The original idea behind it was to use it as a default field sort order, but it's dual-purpose because it also behaves like you said, probably more so, so I renamed it to `IncludeDefaultFields`.

ezekg (approved):

PR #75, opened 2026-02-25
Add renew license action

Cazden:
Adds UI and query for renewing licenses.

API rejection error handling:

Tooltip when license isn't expired:


src/pages/app/license/details.tsx
```diff
@@ -263,45 +288,109 @@ export default function LicenseDetails() {
...
+                          onSelect={(e) => e.preventDefault()}
+                        >
+                          Renew
+                        </DropdownMenuItem>
+                      </TooltipTrigger>
+                      <TooltipContent
+                        side="left"
+                        className="max-w-64 bg-background-4 text-pretty text-content-muted"
+                      >
+                        Renewable upon expiry
```
ezekg: Let's always allow a license to be renewed, even if not expired. Lots of users, including myself, renew before expiration actually occurs (because access might be limited if it does expire).

ezekg (changes requested):
One minor change.

ezekg (approved):

PR #79, opened 2026-02-26
Add license entitlement actions

Cazden:
Adds the endpoints for attach, detach and list license entitlement, as well as entitlement fields for attaching/detaching in license forms.

Adds an Entitlements section to the license page via `list`:

And form fields to the license forms via `attach` and `detach`:


src/components/licenses/form/edit.tsx
```diff
@@ -29,35 +39,98 @@ export default function EditLicenseForm({
   const { data: policy } = useGetPolicy(
     license?.relationships.policy?.data?.id ?? "",
   )
+  const { data: currentEntitlements = [] } = useListLicenseEntitlements(
+    license?.id ?? "",
+  )
 
   const updateLicense = useUpdateLicense(license?.id ?? "")
```
ezekg, line 46: Is there a way to assert license is not null? Seems weird to be on an edit page for a specific license but have to check that the license is not null everywhere. Maybe show a not found error and redirect to index if license is null? Fine to do in a separate PR, since this probably applies to all resources.

ezekg (approved):
One note, but overall looks good!

src/components/licenses/form/edit.tsx
Cazden: Component splitting, maybe? We could do the query in a wrapper component, e.g. `EditLicenseForm` which would query the license, then conditionally render the actual form component `EditLicenseFormContent` based on if the query was successful, and pass it through props:

```
export default function EditLicenseForm() {
  const { data: license, isLoading: licenseLoading } = useLicense(id)

  if (licenseLoading ) return <Loading.Dots /> // Or skeleton
  if (!license) return // Redirect

  return <EditLicenseFormContent license={license} />
}

function EditLicenseFormContent({ license }: { license: License }) {
  const updateLicense = useUpdateLicense(license.id)
  ...
}
```

Otherwise it would probably have to be resolved in the query hook so we don't violate React hook rules, i.e. conditionally calling hooks.

ezekg: I think an early return if `license` is null would let TS know that `license` is not null, no need for splitting.

```ts
const { data: license } = useGetLicense(id)

// after this assert we know license is not null
if (!license) return

const { data: policy } = useGetPolicy(
  license.relationships.policy?.data?.id ?? "",
)

const updateLicense = useUpdateLicense(license.id ?? "")
const form = useForm<Schemas.Licenses.UpdateValues>({
  resolver: zodResolver(Schemas.Licenses.UpdateSchema),
  mode: "onChange",
  values: {
    name: license.attributes.name ?? "",
    expiry: license.attributes.expiry
      ? license.attributes.expiry.slice(0, 16)
      : null,
    suspended: license.attributes.suspended ?? false,
    protected: license.attributes.protected ?? false,
    maxMachines: license.attributes.maxMachines ?? null,
    maxProcesses: license.attributes.maxProcesses ?? null,
    maxUsers: license.attributes.maxUsers ?? null,
    maxCores: license.attributes.maxCores ?? null,
    maxUses: license.attributes.maxUses ?? null,
    metadata: license.attributes.metadata ?? {},
  },
})
```

Still have the weird nullable policy, but that's a separate type issue.

Cazden: It would still have to be split in your example. Hooks can't be called conditionally, so anything after `if (!license) return` would give an error, because `useGetPolicy`, `useForm` etc. are all hooks and their initialization would be conditional on licenses not being null.

https://react.dev/reference/rules/rules-of-hooks

Cazden: Going to mark this as resolved so I can merge, but I've moved it into an issue: #80.

PR #81, opened 2026-03-03
Add license user actions

Cazden:
Adds the endpoints for attach, detach and list license users, as well as user fields for attaching/detaching in license forms. Also includes some refactoring to license/policy entitlement handling that was related.

Shows user count on the license overview sidebar (not sure if you want to link to individual users in the main content area?):

And form fields to the license forms via attach and detach (consolidated into the entitlements step, which was renamed to `Relationships`):


ezekg (approved):
Looks good. New form primitives really shine. Would be nice to open an issue for creating users inline, since all we need is an email to set up a "managed" user.

PR #82, opened 2026-03-04
Add checkout machine action

Cazden:
Adds API-integrated UI for checking out machines, with the same wizard flow that checkout license has.


src/components/machines/form/check-out.tsx
```diff
@@ -0,0 +1,370 @@
...
+      )}
+
+      {showResult && (
+        <SecretModal
+          value={certificate}
+          open={open}
+          onOpenChange={handleOpenChange}
+          title="Checkout summary"
+          warning={
+            "Once this modal is closed, the machine file cannot be read again.\nPlease download the file and store it in a safe location before continuing."
```
ezekg, line 359: Nit: you could use a template string here instead of a newline literal.

ezekg (approved):
Looks good. One minor nit, but feel free to ignore, not a big preference either way.

PR #83, opened 2026-03-04
Add reset machine heartbeat action

Cazden:
Adds API query and UI for resetting machine heartbeats via a simple button that shows a confirmation modal.


src/pages/app/machine/details.tsx
```diff
@@ -643,6 +682,17 @@ export default function MachineDetails() {
         onOpenChange={(value) => toggleOpen("edit", value)}
       />
 
+      <ConfirmationModal
+        title={`Reset heartbeat for ${machine?.attributes.name || machine?.attributes.fingerprint}`}
+        description="Are you sure you want to reset and stop the heartbeat monitor for this machine? This will not deactivate the machine."
```
ezekg: Nit: this message could be improved by adjusting the message based off 1) the machine's heartbeat requirements, and 2) the policy's heartbeat duration.

1. When heartbeat is required, duration is 1 hour: "Are you sure you want to reset the heartbeat monitor for this machine? Because the policy requires a heartbeat, this will cause the machine to be deactivated automatically after 1 hour unless the machine begins sending a heartbeat again."
2. When heartbeat is not required, leave as-is.

ezekg (dismissed):
One minor nit i.r.t. wording, but fine to ignore and open up an issue for later tbh.

ezekg (approved):

PR #89, opened 2026-03-06
 Add change group query to machine forms

Cazden:
Adds the API query for changing the machine group. ~~Will rebase after #88 is merged.~~


src/types/machines.ts
```diff
@@ -116,7 +116,7 @@ export const MachineFormFieldDescriptions: Readonly<
     "This can be an arbitrary string, but must be unique within the scope of the license it belongs to.",
   license: "The license that the machine is for.",
   group: "The group the machine belongs to.",
-  owner: "The user that owns the machine.",
+  owner: "The users that own the machine.",
```
ezekg: Old description was right

ezekg (changes requested):

ezekg (approved):

Issue #61, opened 2026-01-27
Implement pagination for list pages


ezekg:
Right now pagination is non-functional. I'd like to see the ability to set page size, and navigate between pages.

<img width="2704" height="901" alt="Image" src="https://github.com/user-attachments/assets/70bf2ed5-4cf9-4781-8435-860511433a8d" />

PR #92, opened 2026-03-10
Add server-side pagination to list pages

Cazden:
Refactors pagination across all list pages from client-side to server-side ~~and dynamically calculates page size based on available container height/row height, so that lists fill the viewport regardless of screen size/mobile~~ *removed dynamic sizing in favor of static values*.

Also adds a `useDataTable` hook that manages page state, as well as a `Pagination` component for pagination buttons and `PageFooter` for layout, similar to `PageHeader`.

https://github.com/user-attachments/assets/fd20ec8a-dd86-4411-a3a6-d3d0fc4048ae

### DIFF
--- a/src/components/data-table.tsx
+++ b/src/components/data-table.tsx
@@ -1,15 +1,13 @@
-import { useState, useEffect } from "react"
-
+import { useState, useEffect, useRef, useLayoutEffect } from "react"
 import {
   flexRender,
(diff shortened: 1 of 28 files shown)


Cazden:
Also noting this closes #61.

ezekg (approved):
Looks good. LGTM.

PR #93, opened 2026-03-12
Refactor relationship fields across forms to use search inputs

### LINKED ISSUE #91: Update all relationship fields to use search
Follow up to #63, still need relationship fields for products, policies, entitlements, etc. to use search as well.

Cazden:
Closes #91. Finishes what #90 started by refactoring the rest of the form relationship fields with searchable input components that list options and filters search results via the `/search` endpoint. Adds a new search component library that includes refactoring the old `search-select.tsx` to `/search/select.tsx`, adds a `grouped-select.tsx` component that categorizes options (e.g. listing policies by product), and a `multi-select.tsx` component that's allows users to search and select multiple options.


src/components/search/select.tsx
```diff
@@ -182,7 +72,11 @@ export default function SearchSelect<T extends BaseOption>({
   const labelRef = useRef(new Map<string, string>())
 
   const [searchType, searchQuery, searchOp] = useMemo<
-    [ResourceType | null, Record<string, string>, ("AND" | "OR") | undefined]
+    [
+      SearchableResource | null,
+      Record<string, string>,
+      ("AND" | "OR") | undefined,
```
ezekg: Since this is repeated so often, should extract to a shared tuple type.

src/lib/search.tsx
```diff
@@ -0,0 +1,172 @@
...
+    "attributes" in option &&
+    typeof (option as NamedSearchOption).attributes?.name === "string"
+  ) {
+    return (option as NamedSearchOption).attributes.name
+  }
+  return option.id
+}
+
+interface ResourceConfig {
+  getLabel: (option: SearchOption) => string
```
ezekg: Should this be optional, e.g. `getLabel?: ...`? https://github.com/keygen-sh/keygen-portal/pull/93/changes#diff-8cecf23792cec9bf7e60329220ab92e0d3de2218813d3d1ef3cbe98928999757R64

src/lib/search.tsx
```diff
@@ -0,0 +1,172 @@
...
+      </span>
+    ),
+    searchQuery: (term) => ({
+      query: { id: term, name: term, code: term },
+      op: "OR",
+    }),
+  },
+  products: {
+    getLabel: (option) =>
+      (option as NamedSearchOption).attributes?.name ?? option.id,
```
ezekg: Could probably define a `getNamedLabel` function instead of repeating this:

```ts
type NamedSearchOption = SearchOption & { attributes: { name: string } }

function getNamedLabel(option: NamedSearchOption): string {
  return option.attributes.name ?? option.id
}
```

src/lib/search.tsx
```diff
@@ -0,0 +1,172 @@
...
+import GoToButton from "@/components/go-to-button"
+
+type NamedSearchOption = SearchOption & { attributes: { name: string } }
+
+export function getDefaultLabel<T extends SearchOption>(option: T): string {
+  if (
+    "attributes" in option &&
+    typeof (option as NamedSearchOption).attributes?.name === "string"
+  ) {
+    return (option as NamedSearchOption).attributes.name
```
ezekg: This seems verbose, and not a big fan of the type casting to check for properties. You could probably just default to `id` and let resources explicitly opt into using other attributes like you're already doing.

```ts
export function getDefaultLabel(option: SearchOption): string {
  return option.id
}
```

ezekg (changes requested):
In addition to the comments, while testing, I noticed a few issues:

1. Creating entitlements inline during license creation results in duplicated requests, attempting to create the same entitlements multiple times, resulting in the form submission failing, but still creating the entitlements. Not sure what's causing the duplicate requests.
2. When attaching a user to a license using the multi-select search component, filtering down to a user and pressing enter makes the wizard go back a step instead of selecting the user. The other search selects works fine.
3. In addition to above, arrow keys can't be used to navigate multi-select search.

Let me know if you're unable to reproduce.

src/lib/search.tsx
```diff
@@ -32,141 +39,152 @@ interface ResourceConfig {
...
 }
 
+export const licensesConfig: ResourceConfig<License> = {
+  getLabel: getLicenseLabel,
+  placeholder: "Select a license...",
+  searchPlaceholder: "Search by ID or name...",
+  emptyMessage: (
+    <span className="flex items-center gap-2">
+      No licenses found.
+      <GoToButton
```
ezekg, line 53: For another PR, but I think we should prompt the user to confirm they want to navigate away if a form is in progress so that they don't lose data, or we should open the resource list in a modal.

src/components/search/grouped-select.tsx
```diff
@@ -0,0 +1,260 @@
...
+                      {group.label}
+                    </div>
+                    {group.options.map((option) => {
+                      const label = getLabel(option)
+                      const displayOptionLabel = format ? format(label) : label
+
+                      return (
+                        <CommandItem
+                          key={option.id}
+                          onSelect={() => handleSelect(option.id)}
```
ezekg, line 241: When hovering over 2 items with the same label, both are highlighted. Adding `value={option.id}` fixes the issue, since this is [by default inferred from `children`/`textContent`](https://github.com/dip/cmdk/blob/main/cmdk/src/index.tsx#L51-L54).

Image

This applies to all `Command` inputs.

ezekg (changes requested):
One minor change, and then a note about a UX issue that we can tackle later.

src/lib/search.tsx
Cazden: Sounds good, opened an issue for this: #95

ezekg (approved):

PR #96, opened 2026-03-13
Fix issue with disabled pointer events after dialog unmount

ezekg:
Fixes the long-standing issue where the body becomes unresponsive due to `pointer-events: none` sticking around after a dialog is closed. I was able to reproduce this after #91 was merged, as navigating into a create modal, clicking a search dropdown, and then navigating to a "View" page would consistently cause the issue.

This issue is due to the way [`@radix-ui/react-dismissable-layer` handles internal context](https://github.com/radix-ui/primitives/blob/22473d16404bfd446305db5b6c9308aece99fdec/packages/react/dismissable-layer/src/dismissable-layer.tsx#L19-L23). Having multiple versions of the package as peer dependencies will result in state conflicts when different components try to maintain control of the `body`'s pointer events, which was the case for our dependencies on `@radix-ui/react-dropdown-menu`, `@radix-ui/react-select`, `@radix-ui/react-tooltip`, `@radix-ui/react-dialog`, and `@radix-ui/react-popover`.

Not sure if there's a better way to handle this so that we don't become out of sync again. We might be able to use a peer dependency override, but maybe just being aware of the cause is enough for now.


pnpm-lock.yaml
```diff
@@ -1055,19 +1055,6 @@ packages:
       '@types/react':
         optional: true
 
-  '@radix-ui/react-dismissable-layer@1.1.10':
```
ezekg, line 1058: This was the cause btw, having both `1.1.10` and `1.1.11` as peer dependencies, each [with their own context](https://github.com/radix-ui/primitives/blob/22473d16404bfd446305db5b6c9308aece99fdec/packages/react/dismissable-layer/src/dismissable-layer.tsx#L19-L23).

Cazden (approved):

PR #97, opened 2026-03-18
Add horizontal table scrolling

### LINKED ISSUE #47: Add horizontal table scroll
In some cases, and for mobile in particular, tables can be too wide and the content won't fit the screen. Current implementation is to use a `hideOnMobile` prop on `DataTable` to select which attributes are hidden on smaller screens, but this isn't a robust solution because sometimes we need to see those attributes. To properly fix this, tables should be able to scroll horizontally.

> ... I think adding horizontal scroll is fine, and there's no way to avoid it for e.g. licenses with long names, long policy names, etc. so might as well assume it'll happen. I always thought [this scroll behavior](https://x.com/pontusab/status/1992162748134989999) with the arrows was rad.

_Originally posted by @ezekg in https://github.com/keygen-sh/keygen-portal/pull/46#discussion_r2668742882_

Cazden:
Adds horizontal scrolling for tables that exceed screen width, which replaces existing functionality of conditionally showing columns via `hideOnMobile`. This allows us to endlessly expand tables with columns to display more than just a few resource attributes.

Closes #47.

Table with no overflow (not much change except column borders and disabled scroll buttons):

Start of table with overflow (narrowed the above table just to showcase scroll):

End of table:

### DIFF
--- a/src/components/data-table.tsx
+++ b/src/components/data-table.tsx
@@ -1,15 +1,21 @@
-import { useState, useEffect, useRef } from "react"
+import {
+  useState,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useMemo,
+  useCallback,
+} from "react"
 import {
   flexRender,
   SortingState,
   useReactTable,
   getCoreRowModel,
   getSortedRowModel,
-  getPaginationRowModel,
 } from "@tanstack/react-table"
 
+import { Button } from "@/components/ui/button"
 import {
-  Table,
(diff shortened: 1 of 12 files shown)


Cazden:
Added a left-side gradient for when there's columns scrolled past on the left (starts at the last static column, `ID` in this example):

ezekg (approved):
Looks sick. Pushed a couple commits related to #98.

PR #98, opened 2026-03-18
Fix unnecessary use effect

ezekg:
This helps reduce re-renders and align with [best practices](https://react.dev/learn/you-might-not-need-an-effect).

### DIFF
--- a/src/components/calendars/date-picker.tsx
+++ b/src/components/calendars/date-picker.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from "react"
+import { useState, useRef } from "react"
 import { ChevronLeftIcon, ChevronRightIcon } from "lucide-react"
 
 import { cn } from "@/lib/utils"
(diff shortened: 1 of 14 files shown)


src/components/calendars/date-picker.tsx
```diff
@@ -21,10 +21,6 @@ export default function DatePickerCalendar({
 }: DatePickerCalendarProps) {
   const [internalDate, setInternalDate] = useState<Date | undefined>(selected)
 
-  useEffect(() => {
-    setInternalDate(selected)
```
ezekg, line 25: Using `key` here instead https://github.com/keygen-sh/keygen-portal/pull/98/changes#diff-7ceb3a5af369c80ed1fe1c420d5d6be77c625764a038ca9110aaed0009cc0a41R469

Cazden (approved):
LGTM.

PR #105, opened 2026-03-19
Add reset form on close

### LINKED ISSUE #34: View/form states should be reset when a modal is closed
e.g. reopening a form modal shouldn't show partially filled fields from before. Similarly, reopening environments modal should always show the list view, not a details view if that's where a user was last at.

ezekg:
Closes #34. Introduces a form provider component to replace direct usage of `Form` from `react-hook-form`. Also tries to separate "close" from "back" to prevent a user from accidentally exiting a form and losing progress.


Cazden (approved):
Usage with the `Forms` component API looks nice, LGTM.

PR #114, opened 2026-03-26
Add Artifacts UI and API integration

Cazden:
Adds all CRUD endpoints for artifacts, along with some new pattern previews: `advanced-dialog.tsx` is in `/artifacts/dialog/advanced.tsx`.. I did this because I've added a second dialog for artifacts for uploading, `/artifacts/dialog/upload.tsx`, so this keeps them grouped together since they're both dialogs, and this pattern will be implemented in the future for other resources just to keep things consistent (#115). Additionally, `all-attributes.tsx` has been named just `attributes.tsx`, which is, again, setup for a later refactor where I'd like to refactor every resource's `all-attributes.tsx` to be like `fields.tsx`, where we define what fields we want in `include`/`exclude` props, should would be accessed like `Artifacts.Attributes include={[...]} />` (#116).

All of this leaves us with a nice looking folder and file structure imo:

Here's some previews of artifacts:


src/keygen/artifacts/get.ts
```diff
@@ -0,0 +1,23 @@
...
+  id: string
+}
+
+export default async function get({ id }: GetProps): Promise<ArtifactResponse> {
+  const result = (await client.request(
+    `/accounts/${config.id}/artifacts/${id}`,
+    {
+      method: "GET",
+      headers: {
+        Prefer: "no-redirect",
```
ezekg: Could we use `no-download` here so that the API doesn't count this as a download event?

src/keygen/artifacts/create.ts
```diff
@@ -0,0 +1,38 @@
...
+      attributes: compact(attributes),
+      relationships,
+    },
+  }
+
+  const result = (await client.request(`/accounts/${config.id}/artifacts`, {
+    method: "POST",
+    body: JSON.stringify(body),
+    headers: {
+      Prefer: "no-redirect",
```
ezekg, line 33: Observant one, you are. I was wondering if you'd figure this one out lol. gg

ezekg (changes requested):
One minor change, but other than that, looks great.

ezekg (dismissed):

ezekg (approved):

PR #128, opened 2026-03-28
Add filter bar primitives and license filters

### LINKED ISSUE #125: Paginating briefly shows previous page's data while next page is loading
I believe this is due to the `keepPreviousData` function: https://github.com/keygen-sh/keygen-portal/blob/ec23a4b509bf72506ec9780cb9b0899e399c4a89/src/queries/licenses.ts#L53

ezekg:
Closes #125. Adds filtering to licenses and along with it a filter bar component with filter primitives.

Image

Image

Image

Image

After this is merged, I'll do a separate PR for each of the other resources.


src/components/motion/slide.tsx
```diff
@@ -25,17 +25,14 @@ export default function MotionSlide({
...
     }),
     center: {
       [axis]: 0,
       opacity: 1,
-      pointerEvents: "auto" as const,
     },
     exit: (d: 1 | -1) => ({
       [axis]: d * -offset,
       opacity: 0,
-      pointerEvents: "none" as const,
```
ezekg, line 38: This was causing issues with popovers where the table content behind the popover was still interactive, because Motion will animate to the opposite of what we set here, i.e. `pointer-events: none => auto`.

src/components/duration-input.tsx
```diff
@@ -72,6 +72,44 @@ const DefaultPresets: Preset[] = [
...
+  const mapping = ISO_UNIT_MAP[unit.key]
+  if (mapping) return `${mapping[0]}${n}${mapping[1]}`
+  return `PT${seconds}S`
+}
+
+// Parses ISO 8601 duration (e.g. P30D, PT1H, P2W) to seconds
+const ISO_PATTERN =
+  /^P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?$/
+
+function isoToSeconds(iso: string): number | null {
```
ezekg: Would prefer to pull in a package for all this ISO stuff tbh because it all feels error-prone.

src/components/duration-input.tsx
```diff
@@ -119,16 +155,35 @@ interface DurationInputProps {
...
+
+type DurationInputProps = DurationInputSecondsProps | DurationInputIsoProps
+
+export default function DurationInput(
+  props: DurationInputProps,
+): React.ReactElement {
+  const { units, presets, disabled, autoFocus, disabledTooltip, className } =
+    props
+
+  const isIso = props.format === "iso"
```
ezekg: I think this is leftover from me playing with the idea of the duration component having a format, e.g. numeric and ISO, where it sets its value as a numeric duration in seconds or an ISO duration string, respectively. I don't think this is actually being used anymore so might need to revert these changes, as it complicates things.

src/hooks/use-filter-state.ts
```diff
@@ -0,0 +1,85 @@
...
+export function useFilterState<T>(
+  value: T | undefined,
+  defaultValue: T,
+  onChange: (value: T | undefined) => void,
+) {
+  const { remeasure } = useContext(FilterBarContext)
+  const [isDraft, setIsDraft] = useState(false)
+  const [draftValue, setDraftValue] = useState<T>(defaultValue)
+  // Snapshot of value when draft was entered — draft auto-invalidates
+  // when value changes externally (e.g. "clear all"), no manual reset needed.
```
ezekg: This comment could be clearer

src/hooks/use-filter-state.ts
```diff
@@ -0,0 +1,85 @@
...
+      setDraftSnapshot(value)
+      setIsDraft(true)
+      remeasure()
+    },
+    [value, remeasure],
+  )
+
+  return {
+    filterState,
+    currentValue,
```
ezekg: I don't like the property name here, something like `filterValue` might be clearer, or even `value` and then allow the caller to deconstruct the value into whatever name they want if that conflicts with another var.

Same applies to `filterState`. Drop the `filter` prefix, it's cleaner that way.

src/keygen/licenses/list.ts
```diff
@@ -25,6 +45,63 @@ export default async function list({
   if (pageSize != null) {
     params.set("page[size]", pageSize.toString())
   }
+  if (filters != null) {
```
ezekg: Would be cleaner to use safe navigation instead, i.e. `?.`.

src/lib/temporal.ts
```diff
@@ -0,0 +1,17 @@
+const ISO_DURATION_PATTERN =
+  /^P(?:(\d+)Y)?(?:(\d+)M)?(?:(\d+)W)?(?:(\d+)D)?(?:T(?:(\d+)H)?(?:(\d+)M)?(?:(\d+)S)?)?$/
+
+export function isoToHumanDuration(iso: string): string {
```
ezekg: I would not be opposed to pulling in a package for this, but not a huge deal.

src/pages/app/license/list.tsx
```diff
@@ -37,7 +49,7 @@ export default function LicensesList() {
   const [open, setOpen] = useState(false)
 
   return (
-    <section className="flex h-screen flex-col">
+    <section className="flex h-screen flex-col overflow-hidden">
```
ezekg: Needed? I'm not sure, could just be a remnant of trying times.

src/queries/components.ts
```diff
@@ -49,7 +44,6 @@ export function useListComponents(params?: { page: number; pageSize: number }) {
 
       return response
     },
-    placeholderData: params ? keepPreviousData : undefined,
```
ezekg, line 52: I nixed this everywhere because it was the cause of #125.

src/queries/licenses.ts
```diff
@@ -34,14 +29,42 @@ export function useGetLicense(licenseId: string) {
   })
 }
 
-export function useListLicenses(params?: { page: number; pageSize: number }) {
+export type LicenseFilters = {
```
ezekg: This is duplicated: https://github.com/keygen-sh/keygen-portal/pull/128/changes#diff-ccb2275c698b808a95f12f15cc24cdac4c24fb4cb04aef21e820831753b779a1R7

src/queries/utils.ts
```diff
@@ -1,6 +1,30 @@
+import { type QueryKey } from "@tanstack/react-query"
+
 import { partitionFulfilled } from "@/lib/partition"
 import { APIError } from "@/types/api"
 
+// similar to Tanstack's keepPreviousData except it keeps previous data during pagination
+// but clears it when filters change (so the loading skeleton is shown immediately)
+export function keepPreviousDataUnlessRefiltered<T>(nextFilters: unknown) {
```
ezekg: I don't think this is even needed anymore since we're no longer using `keepPreviousData`.

src/queries/utils.ts
```diff
@@ -1,6 +1,30 @@
...
+  return (
+    prevData: T | undefined,
+    prevQuery: { queryKey: QueryKey } | undefined,
+  ) => {
+    const prevKey = prevQuery?.queryKey ?? []
+    const lastKey = prevKey[prevKey.length - 1] // query options e.g. env, filters, etc.
+
+    if (typeof lastKey === "object" && "filters" in lastKey!) {
```
ezekg: Believe it or not, `null` is `typof "object"`, so this could throw.

ffs javascript

src/components/filter-bar/duration-filter.tsx
```diff
@@ -0,0 +1,102 @@
...
+export default function DurationFilter({
+  label,
+  icon,
+  ops,
+  durations,
+  value,
+  onChange,
+}: DurationFilterProps) {
+  const defaultValue = useMemo(
+    () => ({ [ops[0].value]: durations[0].iso }),
```
ezekg: This will throw if `ops` is empty

src/components/filter-bar/duration-filter.tsx
```diff
@@ -0,0 +1,102 @@
...
+    filterState,
+    currentValue,
+    handleActivate,
+    handleConfirm,
+    handleRemove,
+    handleChange,
+  } = useFilterState(value, defaultValue, onChange)
+
+  const currentOp = Object.keys(currentValue)[0]
+  const currentVal = Object.values(currentValue)[0]
```
ezekg: This would be safer as `currentValue[currentOp]` (this pattern occurs a few times in the filter primitives)

src/components/licenses/filter-bar.tsx
```diff
@@ -0,0 +1,219 @@
...
+        ops={ACTIVITY_FILTER_OPS}
+        durations={DURATION_PRESETS}
+        value={filters.activity}
+        onChange={(activity) => onChange({ ...filters, activity })}
+      />
+      <Filters.BooleanFilter
+        label="Activated"
+        icon={Power}
+        value={filters.activated}
+        onChange={(v) => onChange({ ...filters, activated: v })}
```
ezekg: Use `{activated}` instead of `{activated: v}` (this applies to all filters using `v`)

src/components/filter-bar/filter-segment.tsx
```diff
@@ -0,0 +1,245 @@
...
+                  aria-label="Discard filter"
+                >
+                  <X className="size-3" />
+                </button>
+                <button
+                  type="button"
+                  onClick={onConfirm}
+                  disabled={confirmDisabled}
+                  className={cn(
+                    "inline-flex h-full items-center rounded-r-[3px] bg-background-2/60 px-1 transition-colors outline-none",
```
ezekg: These 2 icons could be a bit closer together so that margins are more consistent between segments.

These buttons are also quite small, so might be worth [expanding their hit box](https://x.com/frederic_ooo/status/1952756133732483543).

tsconfig.app.json
```diff
@@ -4,7 +4,7 @@
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
     "target": "ES2020",
     "useDefineForClassFields": true,
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "lib": ["ESNext", "DOM", "DOM.Iterable"],
```
ezekg, line 7: I changed this because I couldn't use `Array.findLast()` even though it's supported in all major browsers

src/components/filter-bar/date-filter.tsx
```diff
@@ -0,0 +1,154 @@
...
+  Popover,
+  PopoverContent,
+  PopoverTrigger,
+} from "@/components/ui/popover"
+
+import * as Calendars from "@/components/calendars"
+
+import { cn } from "@/lib/utils"
+
+import { FilterState, useFilterState } from "@/hooks/use-filter-state"
```
ezekg: I believe this should be `type FilterType`? Should probably add a linter rule for this tbh.

src/components/filter-bar/duration-filter.tsx
```diff
@@ -0,0 +1,124 @@
+import { type LucideIcon } from "lucide-react"
+
+import { type FilterState, useFilterState } from "@/hooks/use-filter-state"
+import {
+  FilterSegmentGroup,
+  FilterSegment,
+  FilterPopoverSegment,
+  FilterOptionList,
+} from "./filter-segment"
+import { isoToHumanDuration } from "@/lib/temporal"
```
ezekg: Can we replace this with `date-fns` `formatDuration({ seconds })`?

ezekg: Nope: https://github.com/date-fns/date-fns/pull/3151

src/components/filter-bar/boolean-filter.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { useFilterState } from "@/hooks/use-filter-state"
+import {
+  FilterSegmentGroup,
+  FilterSegment,
+  FilterPopoverSegment,
+  FilterOptionList,
+} from "./filter-segment"
+
+const BOOLEAN_OPTIONS = [
```
Cazden, line 11: Codebase doesn't really have a solid convention when it comes to declaring global vars, switching between uppercase and pascal case. Typical convention is uppercase so should probably standardize this later so codebase is consistent. Nothing wrong with this line but making a note for a potential issue, e.g. `/src/components/duration-input.tsx` uses pascal case, so it should probably be updated to uppercase.

Cazden (approved):
UI looks great and code is really clean imo. LGTM!

PR #132, opened 2026-04-06
Add settings UI and API implementation

Cazden:
Adds account, current user, default license/user permissions, two-factor, and team management UI (forms, pages, etc.) and API queries to create a Settings section/category on the sidebar rail.

Still heavily WIP. Needs billing section, general style and layout refinements, proper file structure, proper form usage/schemas among other things.

The following functionality is the endgoal:

| Category    | Purpose                                          | Contents                                                                         |
| ----------- | ------------------------------------------------ | -------------------------------------------------------------------------------- |
| General     | account-level overview                           | current account (id, name, slug), current user (id, name, email)                 |
| Billing     | subscription and payments                        | manage subscription, payment methods, invoices (managing links out to Stripe rn) |
| Team        | internal user management                         | invite/remove teammates, roles, etc. (always show 2FA status for each teammate)  |
| Security    | authentication and access control (current user) | password, 2FA, active sessions, personal tokens                                  |
| Permissions | default behavior for customer’s end-users        | default user/license permissions controlling what end-users can do by default    |
| Developers  | technical configuration and platform settings    | protected mode, pinned API version, account public keys                          |

PR #143, opened 2026-04-13
Add user settings UI and API integration

Cazden:
Separates https://github.com/keygen-sh/keygen-portal/pull/132 for user-related settings. Adds a new page for security settings (currently just `change-password`, `second-factor` to follow in another PR), and an added section for the current user in General settings, along with API integration. Additionally, adds `forgot-password` API support to the `users` details page under Licensing.

Security settings:
* Change password

General settings addition includes a `Current Users` section with:
* Email
* First name
* Last name


src/pages/app/settings/general.tsx
```diff
@@ -100,6 +105,77 @@ export default function General() {
                 </Motion.Resize>
               )}
             </div>
+
+            <div className="flex flex-col space-y-2">
+              <h2 className="font-owners-wide text-lg text-content-loud">
+                Current User
```
ezekg:
```suggestion
                You
```

src/pages/app/settings/general.tsx
```diff
@@ -100,6 +105,77 @@ export default function General() {
...
+              </h2>
+              <p className="font-owners-text text-sm text-content-muted">
+                Manage current user settings and information.
```
ezekg:
```suggestion
                Manage your settings and information.
```

src/queries/users.ts
```diff
@@ -241,3 +241,97 @@ export function useUnbanUser(userId: string) {
...
 }
+
+export function useChangePassword() {
+  return useMutation<
+    User,
+    APIError,
+    { oldPassword: string; newPassword: string }
+  >({
+    mutationFn: async ({ oldPassword, newPassword }) => {
+      const meResponse = await keygen.profiles.me()
```
ezekg, line 260: I think the `me` request needs `{root: true}` so that the current user can be read while in isolated environments. Right now these requests fail in an isolated environment.

src/queries/users.ts
```diff
@@ -241,3 +241,97 @@ export function useUnbanUser(userId: string) {
...
+      const current = getResponse.data
+
+      const changes = diff(
+        current.attributes,
+        values,
+      ) as Schemas.Users.UpdateValues
+
+      if (Object.keys(changes).length === 0) return current
+
+      const updateResponse = await keygen.users.update({
```
ezekg, line 329: This currently fails with a 404 while in an isolated environment. since this user doesn't exist in the env. My recommendation would be to use `{root: true}` in all requests for the current user, e.g. `{root: user === client.user}`, since right now the current user always belongs to the global environment.

ezekg (changes requested):
Looks good, but has a few issues i.r.t. isolated environments. I recently hit these in the current dashboard app when adding environment support, so added some ideas on how to work around it.

src/queries/users.ts
Cazden: Good catch, I need to remember to test in isolated environments.

src/components/forms/layout/sheet.tsx
```diff
@@ -81,6 +76,8 @@ export default function FormsContentSheet<T extends FieldValues = FieldValues>({
     try {
       await submitOnce()
     } catch {
+      // API errors are handled inside submitOnce via handleFormError
```
ezekg: Drop the catch if it's not actually used so that we don't accidentally swallow unexpected errors. Though I'm curious why this change was needed vs what's still in the wizard layout? tl;dr?

ezekg (changes requested):
Seeing a weird issue with the inline form where the abandon dialog pops up twice:

https://github.com/user-attachments/assets/9dc1114b-8c09-4829-9548-228c62d9ebe7

src/components/forms/layout/sheet.tsx
Cazden: Embarrassing. This was leftover Claude code during iterating that I missed during cleanup, and if I'm understanding correctly this would actually create a potential bug. Removed in 11ca9e3.

ezekg (approved):

PR #147, opened 2026-04-14
Remove form guard check on touched fields

ezekg:
There was an additional check on touched fields added in #139 that seems like it introduced this bug where the permissions form is unguarded when navigating away from the dirtied page form:

https://github.com/user-attachments/assets/0f492694-66dc-49f8-bcde-7c6454e5f2f1

Not sure what the original intent was, but I'm not sure it's needed. lmk if I'm missing something, though?

### DIFF
--- a/src/components/forms/guard.tsx
+++ b/src/components/forms/guard.tsx
@@ -29,11 +29,10 @@ export function FormRouteGuard({ children }: { children: React.ReactNode }) {
 //           of the main guard to avoid unnecessary re-renders of its children
 function FormRouteGuardModal() {
   const { control } = useFormContext()
-  const { isDirty, isSubmitting, touchedFields } = useFormState({ control })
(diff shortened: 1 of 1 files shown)


Cazden (approved):
Not important, can be safely removed. Was some WIP from #132 that I unintentionally carried over into #139.

Original intent was to prevent autofill from counting a form as dirty (i.e. `Current password` in the `Change password` form would autofill, therefore dirtying the form without actual user input), `touchedFields` was being used to determine if the user manually dirtied the field vs. autofill, but this was WIP functionality that wasn't working 100% or was dependent on older iterations, and if it's causing a new bug, it can be argued that it's not worth implementing right now or at all, since the solution was more on the UX-side rather than a functional issue.

We can open an issue if you want the UX of not having a form guard trigger strictly from autofill though.

ezekg:
I don't think it's a big enough deal tbh.

PR #151, opened 2026-04-15
Add second factor UI and API integration

Cazden:
Adds the user second-factor section to the Security settings page, which includes a two-step `enable` form for generating a second-factor, followed by authenticator setup. Likewise has a `disable` second factor form that requires an OTP to deactivate.

Also adds `qrcode.react` for rendering QR codes in our case here, and a new `otp-input.tsx` reusable component for our OTP code in our second factor form, as well as refactoring our auth OTP with it.


src/components/users/form/second-factor/enable.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+  const handleSubmit = useCallback(async () => {
+    const valid = await form.trigger()
+    if (!valid) return
+
+    const { password } = form.getValues()
+
+    try {
+      if (orphanedFactor) {
+        await deleteSecondFactor.mutateAsync({ id: orphanedFactor.id })
```
Cazden, line 117: Should add a comment explaining why this is needed.

src/components/users/form/second-factor/enable.tsx
```diff
@@ -0,0 +1,295 @@
...
+
+export default function EnableSecondFactorForm({
+  orphanedFactor,
+  onClose,
+}: EnableSecondFactorFormProps) {
+  const [step, setStep] = useState<Step>("password")
+  const [pendingFactor, setPendingFactor] = useState<SecondFactor | null>(null)
+
+  return (
+    <Motion.Resize layoutKey={step}>
```
Cazden: We're duplicating `<Motion.Resize>` here as it's already being used further up the hierarchy in the Security settings page. Should remove this.

src/components/users/form/second-factor/enable.tsx
```diff
@@ -0,0 +1,295 @@
...
+    }
+  }, [form, orphanedFactor, createSecondFactor, deleteSecondFactor, onCreated])
+
+  return (
+    <Forms.Provider form={form}>
+      <Forms.Container.Page>
+        <div className="flex flex-col">
+          <div className="flex items-center justify-between border-b border-accent p-2">
+            <h2 className="ml-2 text-sm text-content-muted">
+              Enabling two-factor authentication
```
Cazden, line 139: We're essentially building our own form section here rather than using primitives, because using primitives in this instance poses some issues.

Using the existing `Wizard` primitive with associated `Step`s gives us the side effect of being able to nav back to the password step, which is redundant, and going back and forth between password and authenticator setup will just continually create/delete `second-factor`. We'd have to expand `Wizard` to allow omitting/locking once completed for certain steps?

As for `Sheet`, its submit handler calls onClose, which ends the entire form flow and orphans our new `second-factor` as soon as its created, as we never make it to authenticator setup.

Could also be cause for a new layout primitive. Going to leave as-is since this is a one-off case, but if this occurs again may need to look at a new primitive, or expanding existing.

src/components/users/form/second-factor/enable.tsx
ezekg: ~~I don't think this delete is needed, though? You don't really need to recreate the second factor. All you have to do is create it once, then show the QR code until it's enabled. The secret is available to read while it's in a disabled state, i.e. it's not a one-time thing like with tokens.~~

src/components/users/form/second-factor/enable.tsx
ezekg: I think a forward-only wizard makes sense, but agree that it can wait.

src/keygen/users/second-factors/create.ts
```diff
@@ -0,0 +1,34 @@
...
+      ...(password != null ? { password } : {}),
+      ...(otp != null ? { otp } : {}),
+    },
+  }
+
+  const result = (await client.request(
+    `/accounts/${config.id}/users/${userId}/second-factors`,
+    {
+      method: "POST",
+      body: JSON.stringify(body),
```
ezekg, line 29: I think these might need `root`? Requests all fail with a 404 in an isolated environment.

ezekg (changes requested):
The API client's second factor requests need `{root: true}`, but other than that looks great.

ezekg (approved):

PR #154, opened 2026-04-15
Fix list query invalidation for create queries

ezekg:
Some resources didn't invalidate the list query so the new resource didn't show up unless you reloaded the app. Also gets rid of the manual list push for new records, instead relying on the API as the source of truth.


Cazden (approved):
Ah glad you fixed this. Was going to open an issue for it soon. LGTM!

PR #155, opened 2026-04-16
Add team settings UI and API integration

Cazden:
Adds a settings page for inviting and managing teammates, which includes a simple form for inviting a teammate, as well as a list page complete with filtering, for listing internal users, and navigates to user detail pages.


src/pages/app/settings/team.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const { data: account } = useGetAccount()
+  const { data: plan } = useGetAccountPlan(
+    account?.relationships.plan?.data?.id,
+  )
+
+  const totalPages = links?.meta?.pages ?? 1
+  const totalUsers = links?.meta?.count
+
+  const maxAdmins = plan?.attributes.maxAdmins
+  const planName = plan?.attributes.name
```
ezekg: I think this needs safe navigation on `attributes`?

src/components/users/team-filter-bar.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <Filters.FilterBar
+      filterCount={filterCount}
+      onClearAll={() => onChange({ roles: InternalRoles })}
+      pinned={
+        <Filters.EnumFilter
+          label="Status"
+          icon={CircleDot}
+          options={STATUS_OPTIONS}
```
ezekg: Let's remove the status filter for teammates as it's been a common point of confusion to see an "inactive" admin, not realizing that they're only "inactive" because they have no active licenses. Eventually I'll add more useful team-management filters like name or 2FA status.

src/components/users/form/invite.tsx
```diff
@@ -0,0 +1,67 @@
...
+          title="Inviting a teammate"
+          onSubmit={handleSubmit}
+          errorMessage="Failed to send invite"
+          submitLabel="Send Invite"
+          isPending={createUser.isPending || resetPassword.isPending}
+          size="compact"
+        >
+          <Forms.Section.Stacking>
+            <Users.Form.Fields
+              include={["email", "internalRole"]}
```
ezekg: I'd like to allow admins to customize the teammate's permissions in the invite form, with the following as required to be able to use Portal, i.e. make these unable to be disabled with a helpful tooltip as to why:

- `account.read`
- `user.read`
- `user.password.update`
- `user.password.reset`
- `user.second-factors.create`
- `user.second-factors.delete`
- `user.second-factors.read`
- `user.second-factors.update`
- `token.generate`
- `token.read`
- `token.revoke`

Eventually we'll need to make Portal permission-aware as well, e.g. disable certain resources/actions.

ezekg (changes requested):
I'd like to add a permissions field to the invite teammate modal, which is something people have been asking for in the current UI. Right now, they have to send an invite, then go and manually adjust permissions, which is kind of annoying.

src/pages/app/settings/team.tsx
Cazden: I can add it, but I don't think it's technically necessary? `Resource` defines `attributes` as non-optional, so it's guaranteed to exist in this situation. I'm going to mark this as resolved without pushing the safe navigation changes, but feel free to unresolve and request the change if you think otherwise during your re-review.

src/components/users/form/fields.tsx
```diff
@@ -546,6 +557,52 @@ function PermissionsField({
...
+        <FormItem>
+          <Forms.Field.Header
+            label="Permissions"
+            variant={fieldVariant}
+            tooltip={descriptions.permissions}
+          >
+            <MultiSelect
+              value={field.value ?? []}
+              onChange={field.onChange}
+              options={AdminPermissions.map((p) => ({
```
ezekg, line 591: We need to add a wildcard permission option so that it's easier to select all.

Related: was there a regression somewhere? I swear we just added support for wildcard/no permissions...

src/components/users/form/fields.tsx
```diff
@@ -546,6 +557,52 @@ function PermissionsField({
   )
 }
 
+const PORTAL_REQUIRED_OPTIONS = PortalRequiredPermissions.map((permission) => ({
+  label: permission,
+  value: permission,
+  tooltip: "This permission is required for portal access.",
```
ezekg:
```suggestion
  tooltip: "This permission is required for Portal access.",
```

src/components/users/form/invite.tsx
```diff
@@ -0,0 +1,69 @@
...
+          title="Inviting a teammate"
+          onSubmit={handleSubmit}
+          errorMessage="Failed to send invite"
+          submitLabel="Send Invite"
+          isPending={createUser.isPending || resetPassword.isPending}
+          size="compact"
+        >
+          <Forms.Section.Stacking>
+            <Users.Form.Fields
+              include={["email", "internalRole", "internalPermissions"]}
```
ezekg: Can we add optional first and last name fields here too? Might as well, to avoid a subsequent update.

ezekg (changes requested):
Gonna be annoying and ask for another couple changes.

src/hooks/use-team-table-columns.tsx
```diff
@@ -0,0 +1,48 @@
...
+        header: "Email",
+        cell: (info) =>
+          info.getValue() || <span className="text-content-muted">--</span>,
+      }),
+      column.attr("role", {
+        header: "Role",
+        cell: (info) =>
+          info.getValue() || <span className="text-content-muted">--</span>,
+      }),
+      column.attr("created", {
```
ezekg, line 43: Can we add 2FA status here too? It's required when going through SOC2 to have a screenshot of all users, with their 2FA status, so would like to keep that functionality for customers that need it.

src/components/users/form/fields.tsx
Cazden: No regression, afaik this was only discussed and I've been meaning to make a pass over all permission fields to add this feature but have been busy with other tasks. Should've opened a formal issue at the time. I'll do that now and work on that after this PR closes.

src/components/tables/second-factor-cell.tsx
```diff
@@ -0,0 +1,38 @@
...
+  if (!userId) return <ResourceCell isEmpty />
+  return <SecondFactorCellContent userId={userId} />
+}
+
+function SecondFactorCellContent({
+  userId,
+}: {
+  userId: string
+}): React.ReactElement {
+  const { data: secondFactors, isLoading, isError } = useGetSecondFactor(userId)
```
ezekg, line 23: This fails for users outside of the nil environment with a 404, since the `Client`'s second factor requests use `{root: true}`. The best option would probably be to store the current user's `id` on the `Client` like we are for e.g. environment, and then do `{root: client.user === id}` instead.

src/components/tables/second-factor-cell.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  if (isError) {
+    return <Badge variant="outline">Unknown</Badge>
```
ezekg: Not a fan of the outline variant tbh. Stands out too much.

```suggestion
    return <Badge variant="disabled">Unknown</Badge>
```

src/components/tables/second-factor-cell.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  const enabled = secondFactors?.some((sf) => sf.attributes.enabled) ?? false
+
+  return (
+    <ResourceCell isEmpty={!secondFactors} isLoading={isLoading}>
+      <Badge variant={enabled ? "secondary" : "disabled"}>
```
ezekg: Let's do this, but we also need to fix the amber color because it's not working.

```suggestion
      <Badge variant={enabled ? "success" : "warning"}>
```

To fix the amber color, we'd need to also add this to `src/index.css`:

```css
--color-amber: var(--color-brand-amber);
```

Not sure of the semantic line between `--color` and `--color-brand`, tbh.

ezekg (approved):
I pushed up a few changes to tackle a couple remaining things I found. Please do a quick review then LGTM.

PR #158, opened 2026-04-17
Add wildcard and none multiselect options

### LINKED ISSUE #157: Add wildcard and none permission options
> We need to add a wildcard permission option so that it's easier to select all.

_Originally posted by @ezekg in https://github.com/keygen-sh/keygen-portal/pull/155#discussion_r3095971001_

Per comment, need to add wildcard/none as selectable options to this and all other permission multiselect dropdowns.

Cazden:
Closes #157. Adds a couple of props to the `multi-select` component that allows us to add `None` and `*` as options, and implements said props in permission form fields across the app.

* Adds a `includeNone` prop that will set the field to an empty array, which can be useful in our case where sending `[]` to the API for permission-related attributes will allow the API to correctly initialize the attribute with an empty array rather than use default permissions from setting the field to `null`. Also, it's an optional prop so we can omit it in certain cases, such as the permissions field in the invite user form where we require some permissions for Portal access.
* Renames `wildcard` to `includeWildcard` prop for consistency with the above addition and better represent that it's a bool.

### DIFF
--- a/src/components/multi-select.tsx
+++ b/src/components/multi-select.tsx
@@ -33,11 +33,12 @@ interface RequiredOption extends Option {
 }
 
 interface MultiSelectProps {
-  value: string[]
-  onChange: (value: string[]) => void
+  value: string[] | null | undefined
+  onChange: (value: string[] | null) => void
   options: Option[]
+  noneOption?: boolean
   requiredOptions?: RequiredOption[]
-  wildcard?: string
+  exclusiveOptions?: string[]
   placeholder?: string
   disabled?: boolean
   autoFocus?: boolean
@@ -49,15 +50,18 @@ export default function MultiSelect({
   value,
   onChange,
   options,
+  noneOption,
   requiredOptions = [],
-  wildcard,
+  exclusiveOptions = [],
   placeholder = "Choose...",
   disabled,
   autoFocus,
   disabledTooltip,
   className,
 }: MultiSelectProps) {
-  const selected = value ?? []
+  const items = value ?? []
+  const hasNoneOption = !!noneOption && requiredOptions.length === 0
+  const isNoneActive = hasNoneOption && value != null && value.length === 0
   const [query, setQuery] = useState("")
   const [open, setOpen] = useState(false)
 
@@ -83,24 +87,38 @@ export default function MultiSelect({
       ),
     [query, options],
   )
+  const showNoneOption = useMemo(
+    () => hasNoneOption && "none".includes(query.toLowerCase()),
+    [hasNoneOption, query],
+  )
 
-  const setSelected = (next: string[], focus = true) => {
+  const emit = (next: string[] | null, focus = true) => {
     onChange(next)
     setQuery("")
     if (focus) inputRef.current?.focus()
   }
 
+  const exclusiveSet = useMemo(
+    () => new Set(exclusiveOptions),
+    [exclusiveOptions],
+  )
+
+  const toggleNone = (focus = true) => {
+    emit(isNoneActive ? null : [], focus)
+  }
+
   const toggle = (value: string, focus = true) => {
     if (requiredSet.has(value)) return
 
-    const isActive = selected.includes(value)
-    const next =
-      wildcard && value === wildcard
-        ? []
-        : isActive
-          ? selected.filter((v) => v !== value)
-          : [...selected, value]
-    setSelected(next, focus)
+    if (exclusiveSet.has(value)) {
+      emit(items.includes(value) ? null : [value], focus)
+      return
+    }
+
+    const base = items.filter((v) => !exclusiveSet.has(v))
+    const isActive = base.includes(value)
+    const next = isActive ? base.filter((v) => v !== value) : [...base, value]
+    emit(next.length === 0 ? null : next, focus)
   }
 
   const remove = (value: string) => {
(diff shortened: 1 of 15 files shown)


src/components/multi-select.tsx
```diff
@@ -33,11 +33,12 @@ interface RequiredOption extends Option {
...
 interface MultiSelectProps {
-  value: string[]
-  onChange: (value: string[]) => void
+  value: string[] | null | undefined
+  onChange: (value: string[] | null) => void
   options: Option[]
+  noneOption?: boolean
   requiredOptions?: RequiredOption[]
-  wildcard?: string
+  exclusiveOptions?: string[]
```
Cazden: Previously, I was including `noneOption` in here as well, hence the refactor from `wildcard` to an array of `exclusiveOptions`, e.g. `["*", "None"]`, but `None` is no longer included here and has its own dedicated prop. Wondering if this array is still necessary or if it could be reverted to something like `includeWildcard`. I think this currently scales better if we want additional options, but is there really another instance where we'll need another exclusive option besides `wildcard`?

src/components/multi-select.tsx
```diff
@@ -83,24 +87,38 @@ export default function MultiSelect({
       ),
     [query, options],
   )
+  const showNoneOption = useMemo(
+    () => hasNoneOption && "none".includes(query.toLowerCase()),
```
Cazden: We can probably simplify this by moving this logic into `toggle`.

src/components/multi-select.tsx
```diff
@@ -83,24 +87,38 @@ export default function MultiSelect({
...
     if (focus) inputRef.current?.focus()
   }
 
+  const exclusiveSet = useMemo(
+    () => new Set(exclusiveOptions),
+    [exclusiveOptions],
+  )
+
+  const toggleNone = (focus = true) => {
+    emit(isNoneActive ? null : [], focus)
```
Cazden: Same as above [comment](https://github.com/keygen-sh/keygen-portal/pull/158/changes#r3101917411).

src/components/multi-select.tsx
```diff
@@ -33,11 +33,12 @@ interface RequiredOption extends Option {
 }
 
 interface MultiSelectProps {
-  value: string[]
-  onChange: (value: string[]) => void
+  value: string[] | null | undefined
+  onChange: (value: string[] | null) => void
   options: Option[]
+  noneOption?: boolean
```
Cazden: Nit: should rename to `includeNone` or something that befits a boolean value better. `noneOption`, `requiredOptions` and `exclusiveOptions` have similar naming but different types so could be a bit confusing.

src/components/multi-select.tsx
Cazden: Opted to refactor to `includeWildcard` as it's the only use-case at the moment and we can always refactor later, along with `includeNone` so naming is consistent.

src/components/products/form/fields.tsx
```diff
@@ -330,14 +334,22 @@ function PermissionsField({
...
-              options={ProductPermissions.map((p) => ({
-                label: p === "*" ? "*" : p,
-                value: p,
-              }))}
-              wildcard="*"
-              placeholder="Leave blank to use defaults"
+              options={[
+                { label: "*", value: "*" },
+                ...ProductPermissions.map((p) => ({
+                  label: p,
```
Cazden: Following the refactor in 96ace37800e8d727fb5f4ce8d4dee71cbd24a278, this label/value pair is verbose and requires more setup at the callsite. Should be internalized like `includeNone` so usage is consistent with both props.

src/components/users/form/fields.tsx
```diff
@@ -540,13 +543,22 @@ function PermissionsField({
...
-              value={field.value ?? []}
+              value={field.value}
               onChange={field.onChange}
-              options={ProductPermissions.map((p) => ({
-                label: p,
-                value: p,
-              }))}
-              placeholder="Leave blank to use defaults"
+              options={[
+                { label: "*", value: "*" },
```
Cazden: [Comment](https://github.com/keygen-sh/keygen-portal/pull/158/changes#r3102587918).

src/components/users/form/fields.tsx
```diff
@@ -586,12 +598,17 @@ function InternalPermissionsField({
...
             <MultiSelect
-              value={field.value ?? []}
+              value={field.value}
               onChange={field.onChange}
-              options={AdminPermissions.map((p) => ({
-                label: p,
-                value: p,
-              }))}
+              options={[
+                { label: "*", value: "*" },
```
Cazden: [Comment](https://github.com/keygen-sh/keygen-portal/pull/158/changes#r3102587918).

src/components/multi-select.tsx
```diff
@@ -76,31 +82,45 @@ export default function MultiSelect({
     () => new Map(options.map((o) => [o.value, o.label])),
     [options],
   )
+  const normalizedOptions = useMemo(() => {
+    const base = options.slice()
+    if (includeWildcard && !base.some((o) => o.value === "*")) {
+      base.unshift({ label: "*", value: "*" })
```
ezekg: I feel like the `*` label might be confusing for non-technical users.

```suggestion
      base.unshift({ label: "All", value: "*" })
```

ezekg (changes requested):
Minor change to wildcard label, but other than that, looks good.

ezekg (approved):

PR #161, opened 2026-04-17
Fix team back navigation

### LINKED ISSUE #159: The sidebar is wrong when navigating to a teammate and navigating back
When you navigate to the Settings > Team page, then navigate to a teammate and either delete that user or click "Users" in the breadcrumbs, the page goes to the normal Users table view, filtered to the default `user` role (i.e. all teammates are missing, inducing heart atttacks), and the sidebar is still on Settings, exacerbating the user's distress. This is confusing because you're expecting to see teammates, not users, especially when deleting a teammate.

https://github.com/user-attachments/assets/09812ce4-437c-4c23-90ea-50a036262a3b

ezekg:
Closes #159. Refactors back navigation to prefer history, resolving issues where "back" wouldn't always mean "back."

### DIFF
--- a/src/components/back-button.tsx
+++ b/src/components/back-button.tsx
@@ -15,11 +15,10 @@ interface BackProps {
 }
 
 export default function BackButton({
-  path = '..',
   label,
   className,
 }: BackProps): React.ReactElement {
-  const back = useBackNavigate(path)
+  const back = useBackNavigate()
 
   return (
     <div className={cn("group flex h-6 w-fit gap-2", className)}>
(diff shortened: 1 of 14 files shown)


src/components/back-button.tsx
```diff
@@ -1,34 +1,25 @@
...
-  path: string
+  path?: string
   label?: string
   className?: string
-  preferHistory?: boolean
 }
 
 export default function BackButton({
-  path,
+  path = '..',
```
ezekg: We should probably just drop this prop because back should always be back, and if some other path is needed, the `GoToButton` should probably be used instead. This might also introduce issues where, if a `path` was supplied, history is used instead possibly resulting in an incorrect navigation.

Cazden (approved):

PR #162, opened 2026-04-17
Fix default filters on user detail

ezekg:
Refactors where default filters are applied, since the defaults on `/users` would also be applied to child routes like `/users/$id`. This resulted in user detail pages having `?roles=["user"]` appended onto the URL, which triggers my OCD severely.


Cazden (approved):
Your OCD triggers can relax.

PR #163, opened 2026-04-17
Standardize global const vars to uppercase instead of pascal case

### LINKED ISSUE #130: Standardize global vars to uppercase
> Codebase doesn't really have a solid convention when it comes to declaring global vars, switching between uppercase and pascal case. Typical convention is uppercase so should probably standardize this later so codebase is consistent.

_Originally posted by @Cazden in https://github.com/keygen-sh/keygen-portal/pull/128#discussion_r3017790146_

Cazden:
Closes #130.


ezekg (approved):
Be still my OCD. LGTM.

PR #168, opened 2026-04-21
Add filter bar to releases

ezekg:
Adds filtering support for releases.


src/components/filter-bar/enum-filter.tsx
```diff
@@ -23,7 +23,7 @@ export default function EnumFilter({
   value,
   onChange,
 }: EnumFilterProps) {
-  const filter = useFilterState(value, "", onChange)
+  const filter = useFilterState(value, options[0]?.value ?? "", onChange)
```
ezekg, line 26: This was a bug where the filter UI showed the first option as the draft mode default, but if you actually applied it, it did nothing.

Cazden (dismissed):

Cazden (approved):

PR #169, opened 2026-04-21
Add filter bar to artifacts

ezekg:
Adds filtering support for artifacts.

### DIFF
--- a/src/lib/search.tsx
+++ b/src/lib/search.tsx
@@ -13,6 +13,8 @@ import {
   Policy,
   Package,
   Release,
+  Platform,
+  Arch,
 } from "@/types"
 
 import { getUserLabel } from "@/lib/users"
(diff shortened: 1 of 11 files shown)


src/lib/search.tsx
```diff
@@ -240,6 +242,28 @@ export const releasesConfig: ResourceConfig<Release> = {
...
+    op: "OR",
+  }),
+}
+
+export const archesConfig: ResourceConfig<Arch> = {
+  getLabel: getNamedLabel,
+  placeholder: "Select an arch...",
+  searchPlaceholder: "Search by ID or name...",
+  emptyMessage: <span>No arches found.</span>,
+  searchQuery: (term) => ({
```
ezekg, line 261: The API doesn't support searching on these yet, but will add soon: https://github.com/keygen-sh/keygen-api/issues/1115

Cazden (approved):

Issue #145, opened 2026-04-14
Metadata field is broken if metadata contains non-string values


ezekg:
Related to #17. e.g. if you have a license with metadata `{ a: 1 }`, the keyvalue field will throw an error.

ezekg:
Fixed in #172.

PR #172, opened 2026-04-22
Add type support to metadata form fields

### LINKED ISSUE #17: Metadata input should support explicit value types with type selector
> Also need to be sure that numerics and booleans are allowed and handled correctly, i.e. not all values will be strings. The current UI does this implicitly, e.g. if it looks like an integer, it's parsed as such. But this ends up being problematic when a user e.g. wants a numeric string. Instead, I think the metadata input should have a value type selector, which explicitly parses the value as a specific type, and errors if unable.
> 
> This type selector could also be expanded to have support for a nested structure in the future, e.g. object/array value types, though I don't think that should be a priority right now.
> 
> I think these values types should be explicitly supported now:
> 
> - `string` (default)
> - `boolean`
> - `integer`
> - `float`
> 
> Any blank values should be parsed as `null`.

_Originally posted by @ezekg in https://github.com/keygen-sh/keygen-portal/pull/16#discussion_r2392289548_

ezekg:
Closes #17 and #145. Adds a type select field to the metadata key-value form, with support for scalars and JSON. To do this, most resource schemas now have explicit input and output types (vs an identical input/output or a hodge-podge of type augments), which allows us to add additional information/data to fields where needed to be used during validation, before the schema's transform step (e.g. a more sophisticated metadata form field, with full schema-based validation via a `type` property on the metadata key-value pair).

image


src/components/key-value-input.tsx
```diff
@@ -1,14 +1,175 @@
...
+
+  switch (type) {
+    case "string": {
+      // Empty string is a valid metadata value.
+      return null
+    }
+    case "integer": {
+      if (!trimmedValue) return "Value is required"
+      if (!INTEGER_PATTERN.test(trimmedValue)) {
+        return "Must be a valid integer"
```
ezekg: This "error" handling is garbage. We should move this into the Zod schema, or at leasxt return actual errors.

src/components/key-value-input.tsx
```diff
@@ -1,14 +1,175 @@
...
+function detectType(v: unknown): MetaType {
+  if (v === null) return "null"
+  if (typeof v === "boolean") return "boolean"
+  if (typeof v === "number") return Number.isInteger(v) ? "integer" : "float"
+  if (typeof v === "object") return "json"
+  return "string"
+}
+
+function toInputString(v: unknown): string {
+  if (v === null || v === undefined) return ""
```
ezekg: Can be simplified to `v == null`.

src/components/key-value-input.tsx
```diff
@@ -1,14 +1,175 @@
...
+  { value: "string", label: "String" },
+  { value: "integer", label: "Integer" },
+  { value: "float", label: "Float" },
+  { value: "boolean", label: "Boolean" },
+  { value: "null", label: "Null" },
+  { value: "json", label: "JSON" },
+]
+
+// Strict integer: optional sign, digits only.
+const INTEGER_PATTERN = /^-?\d+$/
```
ezekg: Why no exponent support for integer?

src/components/key-value-input.tsx
```diff
@@ -32,36 +193,63 @@ export default function KeyValueInput<
...
+        type: "manual",
+        message: "Metadata contains invalid values",
+      })
+    } else {
+      formContext.clearErrors(name)
+    }
+  }, [hasErrors, formContext, name])
+
+  // Clear the error on unmount so a stale manual error doesn't persist.
+  useEffect(() => {
```
ezekg: Is this needed?

src/components/key-value-input.tsx
```diff
@@ -32,36 +193,63 @@ export default function KeyValueInput<
...
+    for (const row of rows) map.set(row.id, validateRow(row))
+    return map
+  }, [rows])
+
+  const hasErrors = Array.from(rowErrors.values()).some((e) => e !== null)
+
+  // Surface a field-level error on the parent form when any row is invalid so
+  // the form submit is blocked and the user gets a consistent validation
+  // signal. Clear it again once all rows are valid.
+  useEffect(() => {
```
ezekg: Is this needed?

src/components/key-value-input.tsx
```diff
@@ -71,37 +259,80 @@ export default function KeyValueInput<
...
+        // have a constrained value space, or when crossing the json boundary
+        // (where the existing text is unlikely to be meaningful in the new
+        // type).
+        let nextValue = row.value
+        if (type === "boolean") {
+          nextValue = row.value === "true" ? "true" : "false"
+        } else if (type === "null") {
+          nextValue = ""
+        } else if (type === "json" && row.type !== "json") {
+          nextValue = ""
```
ezekg: If previous type was string, we should parse into JSON.

src/components/key-value-input.tsx
```diff
@@ -71,37 +259,80 @@ export default function KeyValueInput<
...
+        } else if (row.type === "json" && type !== "json") {
```
ezekg: If next type is string, that should retain the value.

src/components/key-value-input.tsx
ezekg: Since `metadataValueToType` intentionally doesn't handle undefined (`undefined = string`), leaving this alone so it's clear that we're handling both null and undefined alike.

src/components/policies/form/create.tsx
```diff
@@ -86,7 +86,10 @@ export default function CreatePolicyForm({
 
   const schema = useMemo(
     () =>
-      Schemas.Policies.composePolicySchema<Schemas.Policies.CreateValues>(
+      Schemas.Policies.composePolicySchema<
+        Schemas.Policies.CreateValues,
+        Schemas.Policies.CreateFormValues
```
ezekg: I don't like how this reverses the generic params to output, input. Input, output seems more natural.

src/components/policies/form/fields.tsx
```diff
@@ -777,8 +777,8 @@ function MetadataField({
     return (
       <div className="space-y-2">
         <Skeleton className="h-5 w-48 rounded-sm" />
-        {Object.keys(metadata!).map((key) => (
-          <div key={key} className="flex space-x-2">
+        {(metadata ?? []).map((_, i) => (
```
ezekg: Can this be moved up to the `metadata` definition?

src/schemas/account.ts
```diff
@@ -22,51 +22,43 @@ export type FieldNames = FieldPath<AllValues>
 
 const BaseShape = z.object({})
 
-const BaseRules = (schema: z.ZodType<BaseValues>): z.ZodType<BaseValues> => {
+const BaseRules = <S extends z.ZodTypeAny>(schema: S): S => {
```
ezekg: Can we avoid `z.ZodTypeAny`?

src/schemas/metadata.ts
```diff
@@ -0,0 +1,281 @@
...
+const MetadataPairShape = z.object({
+  id: z.string(),
+  key: z.string(),
+  type: z.enum(METADATA_TYPES),
+  value: z.string(),
+})
+
+export const MetadataPairSchema = MetadataPairShape.superRefine((pair, ctx) => {
+  const trimmedKey = pair.key.trim()
+  const trimmedValue = pair.value.trim()
```
ezekg: Drop the trimmed prefix, it's cleaner.

src/schemas/metadata.ts
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// convert a MetadataPair[] into a Record<string, MetadataValue> for serialization
+function metadataPairsToRecord(
+  pairs: MetadataPair[],
+): Record<string, MetadataValue> {
+  const out: Record<string, MetadataValue> = {}
+
+  for (const pair of pairs) {
+    const trimmedKey = pair.key.trim()
```
ezekg: Drop the trimmed prefix

src/schemas/releases.ts
```diff
@@ -70,38 +71,50 @@ const ProductShape = z.object({
   productId: z.string().min(1, "Product is required"),
 })
 
-const BaseRules = (schema: z.ZodType<BaseValues>): z.ZodType<BaseValues> => {
-  return schema.superRefine((data, ctx) => {
-    const version = semver.parse(data.version)
-    if (!version) {
-      return
-    }
+const versionChannelRefinement = (
```
ezekg: Put this back ffs

src/components/policies/form/fields.tsx
```diff
@@ -777,8 +777,8 @@ function MetadataField({
     return (
       <div className="space-y-2">
         <Skeleton className="h-5 w-48 rounded-sm" />
-        {Object.keys(metadata!).map((key) => (
-          <div key={key} className="flex space-x-2">
+        {metadata.map((_, i) => (
+          <div key={i} className="flex space-x-2">
```
ezekg: Should probably use `row.id`

src/components/users/form/invite.tsx
```diff
@@ -37,7 +37,7 @@ export default function InviteUserForm({
   })
 
   const handleSubmit = async (values: Schemas.Users.InviteValues) => {
-    await createUser.mutateAsync(values)
+    await createUser.mutateAsync({ ...values })
```
ezekg: Revert

src/lib/entitlements.ts
```diff
@@ -8,28 +8,37 @@ import { toast } from "@/lib/toast"
...
 type EntitlementValues = {
   attach?: string[]
-  create?: { name: string; code: string }[]
+  create?: { name: string; code: string; metadata?: Record<string, unknown> }[]
 }
 
-interface SettleCreateEntitlementsProps<T extends FieldValues> {
-  form: UseFormReturn<T>
+interface SettleCreateEntitlementsProps<
+  T extends FieldValues,
```
ezekg: Prefer TFieldValues like other generics

src/schemas/account.ts
```diff
@@ -1,72 +1,57 @@
...
-  defaultUserPermissions?: string[] | null
-  defaultLicensePermissions?: string[] | null
-}
-export type AllValues = CombineFormValues<BaseValues, UpdateValues> &
-  DeveloperValues &
-  PermissionsValues
-
-export type FieldNames = FieldPath<AllValues>
 
 const BaseShape = z.object({})
```
ezekg, line 6: What is this?

src/schemas/licenses.ts
```diff
@@ -101,25 +55,36 @@ const PolicyShape = z.object({
   policyId: z.string().min(1, "Policy is required"),
 })
 
-const BaseRules = (schema: z.ZodType<BaseValues>): z.ZodType<BaseValues> => {
+const CreateShape = BaseShape.merge(PolicyShape)
+
+const UpdateShape = BaseShape.partial()
```
ezekg, line 62: Omit key? See machine fingerprint and product dist strat.

src/schemas/artifacts.ts
```diff
@@ -29,26 +12,43 @@ const BaseShape = z.object({
   arch: z.string().trim().nullable().optional(),
   signature: z.string().trim().nullable().optional(),
   checksum: z.string().trim().nullable().optional(),
-  metadata: z.record(z.unknown()).default({}),
+  metadata: MetadataPairsSchema.optional(),
 })
 
 const ReleaseShape = z.object({
```
ezekg, line 18: Should we standardize on FooRelationshipShape?

src/schemas/metadata.ts
```diff
@@ -0,0 +1,276 @@
...
+  key: z.string(),
+  type: z.enum(METADATA_TYPES),
+  value: z.string(),
+})
+
+export const MetadataPairSchema = MetadataPairShape.superRefine((pair, ctx) => {
+  const key = pair.key.trim()
+  const value = pair.value.trim()
+
+  if (!key && !value && pair.type !== "null" && pair.type !== "boolean") {
```
ezekg: Do we need the type asserts?

src/schemas/metadata.ts
```diff
@@ -0,0 +1,276 @@
...
+
+        return
+      }
+
+      const n = Number.parseFloat(value)
+      if (!FLOAT_PATTERN.test(value) || !Number.isFinite(n)) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          path: ["value"],
+          message: "Must be a valid number",
```
ezekg: Float

src/schemas/policies.ts
```diff
@@ -95,6 +80,7 @@ export const BaseShape = z.object({
     .nullish()
     .transform((value) => (value == null ? value : value === 0 ? null : value)),
   maxUses: z.coerce.number().int().positive().nullish(),
+  maxCores: z.coerce.number().int().positive().nullish(),
```
ezekg, line 83: Separate PR?

src/schemas/policies.ts
```diff
@@ -326,11 +314,31 @@ export const ProductShape = z.object({
...
+export type UpdateValues = z.output<typeof UpdateSchema>
+export type AllValues = CombineFormValues<
+  BaseValues,
+  CreateValues,
+  UpdateValues
+>
+
+export type FieldNames = Exclude<
+  FieldPath<AllValues>,
+  "scheme" | "encrypted" | "entitlements" | "product.id"
```
ezekg, line 340: What is this used for?

src/schemas/policies.ts
```diff
@@ -658,29 +670,33 @@ export function composePolicySchema<T extends BaseValues = BaseValues>(
...
+  const parsed = schema.parse({
+    name: "temp",
+    product: { id: "temp" },
+  }) as T
 
   // Empty strings so form initializes with empty fields
   parsed.name = ""
   parsed.product.id = ""
+  // Reset metadata to the form-input shape i.e. MetadataPair[]
+  parsed.metadata = []
```
ezekg, line 689: Is this even needed?

src/schemas/policies.ts
ezekg: #178

src/schemas/policies.ts
```diff
@@ -434,7 +443,7 @@ export const NodeLockedRules = (
         values.maxMachines === 1,
       {
         path: ["maxMachines"],
-        message: "Non‑floating must be 1 or unlimited",
+        message: "Non-floating must be 1 or unlimited",
```
Cazden, line 446: TIL there's 2 hyphens that look visually identical but are different unicode characters.

Cazden (approved):
Looks great! LGTM.

src/schemas/policies.ts
Cazden: Witnessing the evolution of the EM dash.

ezekg: Ha! My editor flags non-ASCII characters, only reason I caught this.

PR #181, opened 2026-04-28
Add SSO authentication support

Cazden:
Adds SSO authentication support via WorkOS to the Portal. Includes auto-redirect when landing on `/auth/...` while already authenticated, post-redirect error handling (new generic 404 page and an SSO-specific error page), and session lifecycle for cookie-based auth.


src/pages/auth/login.tsx
```diff
@@ -57,19 +57,30 @@ export default function Login() {
...
       auth.setEmail(email)
 
-      switch (code) {
+      switch (err.code) {
         case AuthErrorCode.PasswordRequired:
           void navigate({ to: `/${keygen.config.id}/auth/password` })
           break
+        case AuthErrorCode.SsoRequired:
+          if (err.links?.redirect) {
+            window.location.href = err.links.redirect
```
ezekg: Let's make sure to prompt the user before redirecting so the user isn't confused. e.g. what the current UI does after receiving an `SSO_REQUIRED` error is show this prompt:

image

I'd probably `navigate` to `/auth/sso` and handle it there, rather than duplicate the redirect logic.

src/pages/auth/login.tsx
```diff
@@ -57,19 +57,30 @@ export default function Login() {
...
+            return
+          }
+          void navigate({ to: `/${keygen.config.id}/auth/sso` })
```
ezekg: For SSO, you'll also need to add error handling for the redirect back from WorkOS. Right now, on SSO failure, the API redirects back to `/sso/error` with an error `code` query parameter, e.g. `https://portal.keygen.sh/sso/error?code=SSO_USER_NOT_FOUND`. You don't need to handle specific error codes, but I would display the code for support purposes. All codes are [here](https://github.com/keygen-sh/keygen-api/blob/24547fbd82aedfd670803fcb3f8eb9b825cbd29f/app/controllers/auth/sso_controller.rb), plus a dynamic catch all at the end for [WorkOS-specific codes](https://workos.com/docs/reference/sso/get-authorization-url). 

This kind of breaks your `/auth` namespace here, but I guess you could have an `/sso` namespace separate from `/auth`, but lmk if you'd prefer to change it on the API side, as it's not a very big change.

image

src/pages/auth/login.tsx
```diff
@@ -57,19 +57,26 @@ export default function Login() {
         throw new Error("Service is unavailable.")
       }
 
-      const { code } = errors[0] as { code: AuthErrorCode }
+      const err = errors[0] as {
+        code: AuthErrorCode
+        detail?: string
+        links?: { redirect?: string | null }
```
Cazden, line 62: Not using `links` anymore.

src/pages/auth/login.tsx
Cazden, line 100: Should revert this.

src/pages/auth/sso.tsx
```diff
@@ -13,29 +14,80 @@ import {
   FormMessage,
 } from "@/components/ui/form"
 import { Input } from "@/components/ui/input"
+import ConfirmationModal from "@/components/confirmation-modal"
 
 import * as keygen from "@/keygen"
+import { useAuth } from "@/hooks/use-auth"
+import * as Loading from "@/components/loading"
+import { AuthErrorCode } from "@/types/auth"
```
Cazden: Ugh.

src/pages/auth/sso.tsx
```diff
@@ -13,29 +14,80 @@ import {
...
-    })
+    void fetchSsoRedirect(ssoForm.getValues().username)
   }
 
+  useEffect(() => {
+    if (auth.email) {
+      void fetchSsoRedirect(auth.email)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
Cazden: Don't like the `useEffect` here, seems like a code smell. Having to `eslint-disable` reinforces that notion.

src/pages/auth/sso.tsx
```diff
@@ -13,29 +14,80 @@ import {
...
   })
 
+  async function fetchSsoRedirect(email: string) {
+    setLoading(true)
+    setLocalError(null)
+    auth.setError(null)
+    auth.setEmail(email)
+
+    try {
+      const response = await keygen.authenticate({ email })
```
ezekg: Rather than re-authenticating just to re-fetch the error (which you already have from the login step), just pass it as state when navigating. If the redirect state isn't there, e.g. a user navigates to `/auth/sso` directly, redirect to `/auth/login`. I would personally set this on an SSO provider or something and not e.g. a query parameter, to reduce the risk of an open redirect vulnerability.

Cazden: Was actually already working on this and will have a change out shortly.

Cazden: Ended up removing the `/auth/sso` route. I think it's an unnecessary added layer at this point, given that SSO is determined via an API response and getting that from the Login page or SSO page makes no functional difference, and then you're suggesting the route not be directly nav-able if I understand correctly?
> e.g. a user navigates to `/auth/sso` directly, redirect to `/auth/login`

So I think consolidating the logic into `/auth/login` simplifies the state/context management a ton. Let me know if you disagree though.

ezekg: Yep agree, though that kind of breaks the fact that we have e.g. `/auth/pass` for the other scenarios. Are you planning on eventually refactoring that into the main login route too?

src/lib/sso.tsx
```diff
@@ -0,0 +1,25 @@
+import { useSyncExternalStore } from "react"
+
+let url: string | null = null
+const listeners = new Set<() => void>()
+
+function subscribe(fn: () => void): () => void {
```
Cazden: The contents of this file aren't useful anymore and this is nothing more than leftover code from an earlier implementation that didn't work out. Can consolidate this into local state in `/auth/login` rather than keeping context.

src/pages/auth/sso.tsx
Cazden: Yeah I'd like to give the entire auth flow a form overhaul that brings it in line with the rest of the codebase, but don't want to balloon this PR.

ezekg: Cool, totally fine then. Carry on.

src/components/sidebar/panel.tsx
```diff
@@ -259,6 +265,18 @@ export default function SidebarPanel(): React.ReactElement {
...
 
+  const navigate = useNavigate()
+  const { mutateAsync: logout } = useLogout()
+
+  async function handleLogout() {
+    await logout()
+    void navigate({
+      to: "/$accountId/auth/login",
+      params: { accountId: keygen.config.id },
+      replace: true,
```
Cazden: This can be simplified by moving `navigate` into `useLogout`.

src/queries/auth.ts
```diff
@@ -0,0 +1,17 @@
...
+import { useSession } from "@/hooks/use-session"
+
+export function useLogout() {
+  const queryClient = useQueryClient()
+  const { setUser } = useSession()
+
+  return useMutation({
+    mutationFn: keygen.logout,
+    onSuccess: () => {
+      setUser(null)
```
Cazden, line 15: Setting user to `null` here is duplicating what's already done in `keygen.logout`. We should remove this.

src/components/sidebar/panel.tsx
```diff
@@ -259,6 +265,18 @@ export default function SidebarPanel(): React.ReactElement {
   const { isCloud } = useCloud()
   const { hasUpdate, reload } = useAppVersion()
 
+  const navigate = useNavigate()
+  const { mutateAsync: logout } = useLogout()
+
+  async function handleLogout() {
+    await logout()
```
Cazden: Not how we typically initialize hooks, should look more like `const logout = useLogout()`, then be called like `logout.mutate()`.

src/keygen/client.ts
```diff
@@ -43,15 +52,12 @@ export class Client {
...
     const defaultHeaders = {
       Accept: "application/vnd.api+json",
       "Content-Type": "application/vnd.api+json",
       "Keygen-Version": config.version,
-      ...(options.root
-        ? { Authorization: `Bearer ${this.rootToken ?? ""}` }
-        : this.activeToken
-          ? { Authorization: `Bearer ${this.activeToken}` }
-          : {}),
+      ...(authToken ? { Authorization: `Bearer ${authToken}` } : {}),
```
ezekg: For Cloud, we don't need the authorization header since it's now session-based.

src/providers/session-provider.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const [initializing, setInitializing] = useState(true)
+
+  // Sync to client for API headers
+  const setUser = useCallback((id: string | null) => {
+    keygen.client.setUser(id)
+    setUserState(id)
+  }, [])
+
+  const logoutAndRedirect = useCallback(async () => {
+    await keygen.logout()
```
ezekg, line 29: Does this actually execute? I never see the `DELETE` request show up when I logout.

ezekg: Ah, I think this is because SSO sessions don't have a token... might need to think of a way to revoke sessions directly, but the API doesn't expose sessions directly yet. So I actually don't think this should block this PR, because it works fine for token-based sessions.

src/pages/auth/password.tsx
```diff
@@ -90,7 +95,7 @@ export default function Password() {
       storage.setItem("token", token)
 
       keygen.client.setRootToken(token)
```
ezekg: Need `keygen.client.setTokenId(tokenId)` to fix https://github.com/keygen-sh/keygen-portal/pull/181/changes#r3175303681

src/providers/session-provider.tsx
ezekg: Actually, follow up: logout *is* broken until I reload the page, because the client's `tokenId` isn't set on initial login, i.e. `POST /tokens`, but it *does* get set after `GET /me` is hit, which happens during session restore.

src/pages/auth/login.tsx
```diff
@@ -146,25 +158,21 @@ export default function Login() {
...
-            params={{ accountId: keygen.config.id }}
-          >
-            Sign in with SSO
-          </Link>
-        </Button>
-      </div>
+      <ConfirmationModal
+        open={ssoRedirectUrl !== null}
+        title="Single Sign-On"
+        label="Continue to IdP"
```
ezekg: Nit: auto-focus confirm here so user can hit enter to redirect to IdP.

ezekg (changes requested):
Just FYI, I merged https://github.com/keygen-sh/keygen-api/pull/1117, which resolves an issue with sessions where you could authenticate into an environment, but you couldn't authenticate back out. The fix was to set separate per-environment session cookies. You'll need to pull the latest API changes on master and rebuild the image.

Overall, this PR looks good. I had a few comments. To be honest, I'm not 100% sold on the confirmation dialog for the SSO redirect, I think I'd prefer to have it inline or as a separate step like the password or second factor. It just feels kind of awkward and out of place. I personally like the steps.

src/keygen/logout.ts
```diff
@@ -1,12 +1,27 @@
-import config from "@/keygen/config"
+import client from "@/keygen/client"
+import * as tokens from "@/keygen/tokens"
 
 const STORAGE_KEYS = ["token", "tokenId"]
```
ezekg: For Cloud, ideally, we should no longer store the token value inside of local/session storage. Storing the token ID is fine. Related to https://github.com/keygen-sh/keygen-portal/pull/181/changes#r3175154944.

src/keygen/config.ts
```diff
@@ -1,6 +1,7 @@
 const config = {
   host: import.meta.env.VITE_KEYGEN_HOST,
   mode: import.meta.env.VITE_KEYGEN_MODE,
+  isCloud: import.meta.env.VITE_KEYGEN_MODE === "multiplayer",
```
ezekg: We should follow the API's definition of Cloud: https://github.com/keygen-sh/keygen-api/blob/2a23e05ff4ce1da1856cf1ce5922d7e8163992db/lib/keygen.rb#L61

We should also update the `isCloud` hook to follow that definition too.

src/providers/session-provider.tsx
Cazden: Fixed logout being broken for token auth in da24a05. Sessions remains an outstanding problem since there's not a way to revoke them yet. Created an issue (#183) to update once API supports that though.

src/pages/auth/sso.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const auth = useAuth()
+  const navigate = useNavigate()
+
+  async function onContinue() {
+    setLoading(true)
+    setError(null)
+
+    try {
+      const { errors } = await keygen.authenticate({ email: auth.email! })
```
ezekg: This is an unnecessary request, since we just made it in the `/login` step. Use a context provider or something to share login state between pages instead of making a duplicate request.

Cazden: That's actually what I had originally, but doing it this way fixes an edge-case where if the user lingers on `/sso` after generating the link from `/login`, the link would eventually expire and the user would land on `/sso/error` after attempting to authenticate with the expired link. Generating the link here is an extra API request, yes, but it solves that since it's a brand new link as soon as they hit continue. Fine with reverting, but just wanted to make you aware of why it was done this way.

ezekg: Gotcha. I think if the user waits, I'm fine with them seeing an expired link and having to try again.

Cazden: Sounds good. Fixed in 8bf1be0.

ezekg (approved):
Looks good! LGTM!

Issue #183, opened 2026-05-06
Add revoke session to logout after API supports it


Cazden:
Related to #181 and mentioned [here](https://github.com/keygen-sh/keygen-portal/pull/181#discussion_r3176050595). API doesn't expose sessions yet but once that's supported, will need to update Portal so logout actually revokes the active session. Currently, logging out from a session and then refreshing the page will automatically re-authenticate into the app, so logout is effectively broken for SSO.

PR #185, opened 2026-05-07
Add permission-gating to Portal

### LINKED ISSUE #156: Make Portal permission-aware
> ... Eventually we'll need to make Portal permission-aware as well, e.g. disable certain resources/actions.

_Originally posted by @ezekg in https://github.com/keygen-sh/keygen-portal/pull/155#discussion_r3095097094_

Cazden:
Closes #156, and related to #183. Adds permission gating to Portal in two ways:

1. New components `<Can>` / `<Can.Any>` / `<Can.All>` wrap UI actions (create, edit, delete buttons, etc.) to block rendering if the user lacks permission:
```
<Can permission="policy.update">
  <Button>
    Edit
  </Button>
</Can>
```

2. New utils  `requirePermission` / `requireAnyPermission` / `requireAllPermissions` for cases where we want to block routes if the user lacks permission:
```
export const Route = createFileRoute("/$accountId/app/arches")({
  component: () => <Page.App.Arches />,
  beforeLoad: ({ context }) =>
    requirePermission(context.queryClient, "arch.read"),
})
```

Also, filters the sidebar based on user permissions, so resources that the user doesn't have read permissions for won't be listed, and subsequently if they don't have permissions to read any resources under a sidebar category, the entire category will be hidden. Finally, also blocks end-users (external `User` role) from authenticating into the Portal dashboard.

### DIFF
--- a/src/components/sidebar/panel.tsx
+++ b/src/components/sidebar/panel.tsx
@@ -236,7 +236,7 @@ const VIEWS: View[] = [
         to: "/$accountId/app/team",
         label: "Team",
         params: { accountId: keygen.config.id },
-        requires: ["admin.read", "user.read"],
+        requires: ["user.read"],
       },
       {
         to: "/$accountId/app/permissions",
@@ -277,10 +277,10 @@ function useActiveView(): View {
 }
 
 function useVisibleViews(): View[] {
-  const { canAny } = usePermissions()
+  const { canAll } = usePermissions()
 
   const isRouteVisible = (route: ViewRoute): boolean =>
-    route.requires == null || canAny(route.requires)
+    route.requires == null || canAll(route.requires)
 
   const filtered = VIEWS.map((view) => ({
     ...view,
--- a/src/components/sidebar/panel.tsx
+++ b/src/components/sidebar/panel.tsx
@@ -56,7 +56,11 @@ import * as keygen from "@/keygen"
 import { useCloud } from "@/hooks/use-cloud"
 import { useMobile } from "@/hooks/use-mobile"
 import { useAppVersion } from "@/hooks/use-app-version"
+import { usePermissions } from "@/hooks/use-permissions"
+
 import { useLogout } from "@/queries/auth"
+
+import type { Permission } from "@/types/users"
 
 import Command from "./command"
 import Combobox from "./combobox"
@@ -77,6 +81,7 @@ type ViewRoute = {
   to: string
   label: string
   params: Record<string, unknown>
+  requires?: readonly Permission[]
 }
 
 type View = {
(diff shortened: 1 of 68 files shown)


src/components/sidebar/panel.tsx
```diff
@@ -251,8 +276,29 @@ function useActiveView(): View {
   return VIEWS.find((view) => view.id === ViewId.Home)!
 }
 
+function useVisibleViews(): View[] {
+  const { canAny } = usePermissions()
+
+  const isRouteVisible = (route: ViewRoute): boolean =>
+    route.requires == null || canAny(route.requires)
```
ezekg: I think this should be `canAll`?

Cazden: Yeah `canAll` makes more sense. Originally was `canAny` because the Teams page's logic was either/or, but that doesn't really make sense. If a user has `admin.read` but not `user.read`, it wouldn't work anyway. I reduced the permission requirement to view Teams to `user.read` and changed this to `canAll`, and the Team page now has some conditional logic in case the user is lacking `admin.read`.

src/providers/session-provider.tsx
```diff
@@ -44,36 +47,44 @@ export function SessionProvider({
...
+        if (!verified) {
           const tokenResource = meResponse.included?.find(
             (r) => r.type === "tokens",
           )
           keygen.client.setTokenId(tokenResource?.id ?? null)
-          setUser(meResponse.data.id)
-          return
         }
+
+        queryClient.setQueryData(["users", "me"], meResponse.data)
```
ezekg: Should this not be wrapped in a `useMe` query like other resources? Seems like a code smell that we're using the `keygen` client directly above and then setting query data directly here.

src/pages/app/settings/team.tsx
```diff
@@ -38,6 +41,16 @@ export default function TeamPage() {
     [table, setFilters],
   )
 
+  // NB(cazden) API requires 'admin.read' whenever an admin appears in the response,
+  // remove it from the request if current user can't read admins to avoid 403.
+  const requestRoles = useMemo(
+    () =>
+      (filters.roles?.length ? filters.roles : InternalRoles).filter(
+        (r) => can("admin.read") || (r as UserRole) !== UserRole.Admin,
```
ezekg, line 49: If the user doesn't have `admin.read`, we can probably just block the team page entirely. Better to not show anything than show inaccurate information.

src/routes/$accountId/app.team.tsx
```diff
@@ -20,4 +21,6 @@ function validateSearch(search: Record<string, unknown>): UserFilters {
 export const Route = createFileRoute("/$accountId/app/team")({
   component: () => <Page.App.Settings.Team />,
   validateSearch,
+  beforeLoad: ({ context }) =>
+    requirePermission(context.queryClient, "user.read"),
```
ezekg: Should probably be `admin.read` and `user.read`.

ezekg (dismissed):
Few notes, but overall looks really good. I'm approving, so feel free to address or not address the comments because they're not really important enough to be blockers.

src/providers/session-provider.tsx
Cazden: Removed all this as the changes to `session-provider` in this PR are mostly no op at this point, was originally related to auth gating end-users but I didn't do a good job at cleanup when I removed that, so it's just vestigial now. Oops.

ezekg (approved):
LGTM!

Issue #192, opened 2026-05-20
Add support for CE


ezekg:
To add support for CE, we need to tackle the following issues:

1. Stop relying on `data.attributes.permissions` in CE.
   * This attribute is not included in CE.
   * Use the default permission set for the role instead.
2. Disable environments.
   * Environments are only available in EE.
3. Plan to disable request logs in CE.
   * This should be handled when logs are added.
4. Plan to disable event logs in CE.
   * This should also be handled when added.

PR #202, opened 2026-05-30
feat: add CE (Community Edition) support

necipsunmaz:
## Summary

Adds initial CE (Community Edition) support as described in #192.

- **Stop relying on `data.attributes.permissions` in CE** — when the `permissions` attribute is absent (null), the provider and `effectivePermissions()` now fall back to a role-based default permission set (`DefaultPermissionsByRole`) instead of returning an empty set.
- **Disable environments in CE** — the `EnvironmentProvider` is split into a CE no-op variant and the existing EE variant; the sidebar environment switcher and "Manage Environments" option are hidden when `VITE_KEYGEN_EDITION !== "EE"`.
- **Request logs / event logs** — not yet implemented; will be gated via `config.isCE` when added, as planned in #192.

## Changes

| File | Change |
|------|--------|
| `env.d.ts` | Added `VITE_KEYGEN_EDITION?` to `ImportMetaEnv` |
| `src/keygen/config.ts` | Added `isCE` flag (`VITE_KEYGEN_EDITION !== "EE"`) |
| `src/hooks/use-edition.ts` | New `useEdition()` hook, mirrors existing `useCloud` pattern |
| `src/types/users.ts` | Added `DefaultPermissionsByRole` — role-based permission sets for CE fallback |
| `src/lib/permissions.ts` | `effectivePermissions()` falls back to role defaults when `permissions` is null in CE |
| `src/providers/permissions-provider.tsx` | CE-aware permission resolution using same fallback logic |
| `src/providers/environment-provider.tsx` | CE variant is a no-op; EE variant preserves existing behavior |
| `src/components/sidebar/combobox.tsx` | Environment switcher hidden in CE |

## Test plan

- [ ] Set `VITE_KEYGEN_EDITION=CE` and start the dev server (`pnpm run dev`)
- [ ] Log in — environment switcher should not appear in the sidebar
- [ ] Navigate to `/environments` if routed — no crash, environments not accessible
- [ ] Verify that users with different roles get the expected permissions via the `can()` helper
- [ ] Set `VITE_KEYGEN_EDITION=EE` — environment switcher and full permission behavior should be unchanged
- [ ] `pnpm lint:check` and `pnpm format:check` pass

### DIFF
--- a/env.d.ts
+++ b/env.d.ts
@@ -5,6 +5,7 @@ interface ImportMetaEnv {
   readonly VITE_KEYGEN_MODE: string
   readonly VITE_KEYGEN_VERSION: string
   readonly VITE_KEYGEN_ACCOUNT_ID: string
+  readonly VITE_KEYGEN_EDITION?: string
 }
 
 interface ImportMeta {
--- a/src/components/sidebar/combobox.tsx
+++ b/src/components/sidebar/combobox.tsx
@@ -20,6 +20,7 @@ import { Environment } from "@/types/environments"
 
 import { useListEnvironments } from "@/queries/environments"
 import { useEnvironment } from "@/hooks/use-environment"
+import { useEdition } from "@/hooks/use-edition"
 
 import * as Environments from "@/components/environments"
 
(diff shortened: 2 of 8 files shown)


Cazden (approved):
Looks great, appreciate the contribution and getting this out of the way. LGTM!

Issue #193, opened 2026-05-21
Add request logs


ezekg:
List view only for now, filterable like other resources.

Issue #131, opened 2026-04-06
Replace offset-based pagination with cursor-based pagination


ezekg:
See https://github.com/keygen-sh/keygen-api/pull/1096. Related to #129. I've deprecated offset-based pagination (`page[number]`) as it has performance and correctness issues and a cursor-based approach (`page[cursor]`) is much more scalable, especially for large datasets.

Docs: https://keygen.sh/docs/api/pagination/

PR #203, opened 2026-06-01
Add event logs

### LINKED ISSUE #194: Add event logs
This includes a general filterable firehose list view, and per-resource event logs in the right sidebar.

Cazden:
Closes #194. Implements Event Logs which includes a list and details view, as well as an Events sidebar per resource.

List view has UTC/local timestamp hover, similar to the license expiration heatmap:

Details view features a new `<Diff />` component for highlighting changes:

Related to #193, resource detail pages now feature a mostly functional Events timeline.

Event logs are also gated behind Enterprise, so I've added a locked overlay/CTA to reflect this:


src/components/event-logs/feed.tsx
```diff
@@ -0,0 +1,291 @@
...
+        You do not have permission to view event logs.
+      </div>
+    )
+  }
+
+  const nextCursor = cursorFromLink(links?.next)
+  const loading = isLoading || isFetching
+  const pageCount = Math.max(
+    1,
+    links?.meta?.pages ?? page + (nextCursor ? 1 : 0),
```
ezekg: Cursor-based pagination doesn't include a page count, because counting is expensive for large tables, so I've opted out of counting for performance reasons. This also means that we don't have a "last page" concept, or a "page count" concept; there's only "next" (and "prev" if the client keeps a cursor history). Rather than show an inaccurate total page count (i.e. current page + 1), don't show anything at all.

Ref: https://github.com/keygen-sh/keygen-portal/issues/131

src/keygen/event-logs/list.ts
```diff
@@ -0,0 +1,71 @@
...
+config.validate()
+
+interface ListProps {
+  limit?: number
+  pageCursor?: string | null
+  pageSize?: number
+  filters?: EventLogFilters
+}
+
+function applyResourceFilter(
```
ezekg: I think a clearer name might be `applyPolymorphicResourceFilter`, since e.g. `request` is also technically a resource, it's just not a polymorphic resource like `resource` (i.e. one that can be many types like `resource[type]=user` or `resource[type]=license`).

src/lib/event-logs.ts
```diff
@@ -0,0 +1,123 @@
...
+        after: pair[1],
+      }
+    }
+
+    if (value && typeof value === "object" && !Array.isArray(value)) {
+      const record = value as Record<string, unknown>
+      return {
+        key,
+        before: record.from ?? record.before ?? record.old ?? null,
+        after: record.to ?? record.after ?? record.new ?? null,
```
ezekg: Image

src/pages/app/artifact/details.tsx
```diff
@@ -399,9 +402,10 @@ export default function ArtifactDetails() {
 
                 {isMobile && (
                   <CollapsibleCard title="Events">
-                    <p className="text-sm text-content-subdued">
-                      Nothing here yet.
-                    </p>
+                    <EventLogs.Feed
+                      compact
+                      filters={{ resource: { type: "artifacts", id } }}
```
ezekg: Prefer singular `type` for polymorphic filters. It's just [happen-chance that this works right now](https://github.com/keygen-sh/keygen-api/blob/b5165d15fa55787b2deda83dc9d9212690883a87/app/models/event_log.rb#L99) (the `classify` method performs `singularize`), but I'd say it's undocumented behavior and shouldn't be relied upon.

src/pages/app/event-log/list.tsx
```diff
@@ -0,0 +1,160 @@
...
+      filters,
+    },
+    { enabled: isEE },
+  )
+
+  const nextCursor = cursorFromLink(links?.next)
+  const loading = isLoading || isFetching
+  const pageCount = Math.max(
+    1,
+    links?.meta?.pages ?? page + (nextCursor ? 1 : 0),
```
ezekg: See: https://github.com/keygen-sh/keygen-portal/pull/203/changes#r3349354587

Basically, fine to say "page 3" to indicate the current page, but not okay to say "page 3 of 4" because "4" will change to "5" and so on, since we don't really know the total page count with cursor pagination.

src/hooks/use-cursors.ts
Cazden, line 1: This is generic enough to be renamed to `use-cursors`, as nothing in here is specific to event logs. Also, this addition would be related to #131.

ezekg (dismissed):
Made a few changes. Take a look, feel free to merge if it looks OK.

ezekg (dismissed):

ezekg (approved):

PR #204, opened 2026-06-02
Add data to license expiration heatmap

Cazden:
Replaces mock data that we were previously using for the license expiration heatmap with real data.


src/queries/licenses.ts
```diff
@@ -70,6 +70,40 @@ export function useListLicenses(
   }
 }
 
+export function useExpiringLicenses(options?: { within?: string }) {
```
ezekg: All charts will have a private analytics endpoint for a data source: https://github.com/keygen-sh/keygen-api/blob/master/features/priv/analytics/heatmaps.feature

ezekg (changes requested):
This is using the wrong data source. You'll need to rewrite it using the private analytics API, which aggregates all of this data much more efficiently on the database side.

src/queries/licenses.ts
ezekg: Then on cell click, you'll want to lazily load expiring licenses for that date:

```
GET /v1/accounts/<account>/licenses?expiring[on]=<date>
```

Ideally with small `limit` and a jump to filtered list view if there's more than a handful on that date.

src/components/analytics/license-expiration-heatmap.tsx
```diff
@@ -225,45 +251,18 @@ export default function LicenseExpirationHeatmap() {
...
+  }, [entries])
 
   return (
     <Chart.Card
       title="License expirations"
-      isLoading={licensesLoading}
+      isLoading={isLoading}
       action={
         <GoToButton
           path="/$accountId/app/licenses"
```
ezekg: Can this go to the pre-filtered view `/licenses?expires={"within"%3A"P1Y"}`?

src/components/analytics/license-expiration-heatmap.tsx
ezekg: After the above, change to "View all."

src/components/analytics/license-expiration-heatmap.tsx
```diff
@@ -687,41 +679,53 @@ function MobileHeatmapGrid({
...
-      {licenses.length > 10 && (
-        <p className="mt-1 text-xs text-content-disabled">
-          +{licenses.length - 10} more
-        </p>
+      {count > PREVIEW_LIMIT && (
+        <GoToButton
+          path="/$accountId/app/licenses"
+          params={{ accountId: keygen.config.id }}
+          search={{ expires: { on: date } }}
+          label={`See all ${count} licenses`}
```
ezekg: Change to "View all" for consistency with the rest of the UI.

src/components/analytics/license-expiration-heatmap.tsx
```diff
@@ -687,41 +679,53 @@ function MobileHeatmapGrid({
...
           <li key={license.id}>
             <GoToButton
               path="/$accountId/app/licenses/$id"
               params={{ accountId: keygen.config.id, id: license.id }}
               label={
-                license.name || truncateKey(license.key, { maxLength: 24 })
+                license.attributes.name ||
+                truncateKey(license.attributes.key, { maxLength: 24 })
               }
               className="[&_button]:truncate [&_button]:text-xs [&_button]:text-content-subdued [&_button]:hover:text-content-loud"
```
ezekg: I think the chevrons should match the text color when not hovered, looks too busy with the teal chevrons:

Image

ezekg: I think this list could use a bit of love, because it makes the list of license the least important thing and the link the most important. Maybe the licenses should be in a table of ID and name, and the "view all" gray?

Fine to do this in a separate PR. Actually, once this is merged, I can play with it.

ezekg: Disregard. I'll redesign this after this is merged because idk what I want yet.

ezekg (approved):

PR #207, opened 2026-06-04
fix: align CE default role permissions with API defaults

### LINKED ISSUE #205: Default permissions are wrong in CE for some roles
The default permissions for support agents and users are wrong:

https://github.com/keygen-sh/keygen-portal/blob/29b76d41f023adc6df13ccda19dcf10168b004f0/src/types/users.ts#L419-L428

Introduced in #202.

necipsunmaz:
## Summary

Fixes #205

In CE, the API does not return a `permissions` attribute, so the Portal falls back to `DefaultPermissionsByRole` to compute the current user's effective permissions. This fallback did not match the defaults the API actually assigns — see `User#has_permissions` in [keygen-api `app/models/user.rb`](https://github.com/keygen-sh/keygen-api/blob/master/app/models/user.rb#L84-L117):

| Role | Before | After (per API) |
|------|--------|-----------------|
| Support agent | `ReadOnlyPermissions` | full permission set (`Permission::ADMIN_PERMISSIONS`) |
| User | `[]` | `Permission::USER_PERMISSIONS` minus the backwards-compat omissions |
| Sales agent | `ReadOnlyPermissions` + curated extras | full permission set |
| Developer | all minus billing | full permission set |
| Read-only | missing 5 permissions | exact `Permission::READ_ONLY_PERMISSIONS` |
| Admin | full permission set | unchanged |

## Changes

- `DefaultPermissionsByRole` now mirrors the API's default lambda for all six roles. Admin, developer, sales agent and support agent share a single `AdminPermissions` reference, mirroring `Permission::ADMIN_PERMISSIONS`.
- `ReadOnlyPermissions` aligned exactly with `Permission::READ_ONLY_PERMISSIONS` — it was missing `account.billing.read`, `token.generate`, `user.password.reset`, `user.password.update` and `user.second-factors.read`. Two of those (`token.generate`, `user.password.update`) are in `PortalRequiredPermissions`, so read-only users were incorrectly locked out of the Portal in CE.
- `UserPermissions` aligned exactly with `Permission::USER_PERMISSIONS` — it was missing `group.licenses.read`, `group.machines.read` and `user.password.reset`. Side effect: these three now also show up as selectable options in the account settings default user permissions field.
- Added `DefaultUserPermissions`, derived as `UserPermissions` minus the permissions the API omits by default for backwards compatibility (`account.read`, `license.users.attach`, `license.users.detach`, `policy.read`, `product.read`).
- Removed now-unused `BillingPermissions`, `SalesAgentPermissions` and `DeveloperPermissions`.

## Notes

- EE/Cloud behavior is unchanged — when the API returns `permissions`, the fallback is never consulted.
- `UserRoleDescriptions` (e.g. "Support Agents … cannot create, update or delete resources") now diverges from the actual role defaults, but since those strings match the documented role semantics — and the API has a FIXME noting role defaults may change — I left them untouched. Happy to update them here if you'd prefer.
- The API also lets `account.settings.default_user_permissions` override the user-role default; the CE fallback doesn't consult account settings (pre-existing behavior from #202). Can follow up separately if desired.

## Test plan

- [x] `pnpm lint:check`, `pnpm format:check`, `tsc --noEmit`, `pnpm build`, `pnpm test` all pass
- [x] Verified each permission list matches the corresponding `Permission::*` constant in keygen-api 1:1 (scripted set comparison)
- [x] With `VITE_KEYGEN_EDITION=CE`, read-only users now satisfy `PortalRequiredPermissions`; users get the API's default user permission set instead of none


Cazden:
https://github.com/keygen-sh/keygen-portal/pull/210#issuecomment-4650829705

PR #209, opened 2026-06-04
fix: replace downloaded SVG files with simple-icons package imports

### LINKED ISSUE #113: Replace simple-icons SVG files with package imports
We currently have SVG files directly downloaded from simple-icons (e.g. platform/engine icons in the releases form). Instead, we should install the simple-icons package and use its icon objects. Would be easier from a maintainability standpoint, e.g. if brand logos/colors change.

necipsunmaz:
Fixes #113

Replaces the manually downloaded SVG files in `src/icons/` with imports from the `simple-icons` package. Brand hover colors now come from the package instead of being hardcoded, so they stay in sync automatically.

Also removes `vite-plugin-svgr` since nothing uses `.svg?react` imports anymore.

Side effect: Tauri's hover color changes from `#FFC131` to `#24C8D8` — that's the current brand color per simple-icons, so it's actually a fix.


Cazden:
https://github.com/keygen-sh/keygen-portal/pull/210#issuecomment-4650829705

PR #211, opened 2026-06-08
Add cursor-based pagination

### LINKED ISSUE #129: Pagination shows 1 of 0 when table is empty
Page count should always be at least 1.

Image

### LINKED ISSUE #131: Replace offset-based pagination with cursor-based pagination

Cazden:
Closes #129, closes #131. Replaces offset-based pagination with cursor-based across all paginated lists, building off what #203 implemented for event logs.

Worth noting that the team page no longer shows a teammate count in the header, since it was derived from the previous offset-based pagination via `meta.count`.

### DIFF
--- a/src/components/artifacts/filter-bar.tsx
+++ b/src/components/artifacts/filter-bar.tsx
@@ -39,19 +39,19 @@ export default function ArtifactFilterBar({
   }, [])
 
   const { data: products = [] } = useListProducts(
-    { page: 1, pageSize: 10 },
+    { cursor: "", pageSize: 10 },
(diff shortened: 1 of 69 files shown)


ezekg (approved):
Looks good. LGTM!

PR #219, opened 2026-06-10
Add webhooks

### LINKED ISSUE #195: Add webhooks
Webhook endpoints and events need to be added.

Cazden:
Closes #195. Adds webhook endpoints and events list and details pages.

Webhook endpoints:

Webhook events:


src/types/events.ts
```diff
@@ -0,0 +1,219 @@
...
+  created: string
+  updated: string
+}
+
+export type EventRelationships = {
+  account: Relationship<Linkage<"accounts">>
+  environment: Relationship<Linkage<"environments"> | null>
+}
+
+export type Event = Resource<
```
ezekg: I think we need to stick to the 'webhook' topic, because we have many "events," and will have more later, e.g. event logs, event types, wehbook events, meter events (future). Maybe you can nest this under a `webhooks` directory if you want to keep the name? But imo, these should align with the API models: `WebhookEndpoint` and `WebhookEvent` so that it's easier to maintain.

src/types/events.ts
```diff
@@ -0,0 +1,219 @@
...
+  status: "The current status of the event.",
+  apiVersion:
+    "The API version that the webhook event payload is formatted for.",
+  lastResponseCode:
+    "The last HTTP response code that your webhook endpoint sent in response to the webhook.",
+  lastResponseBody:
+    "The last HTTP response body that your webhook endpoint sent in response to the webhook.",
+} as const
+
+export const EventTypes = [
```
ezekg, line 1: This should be moved to `event-types` since it's a separate type shared between multiple resources.

src/schemas/webhook-endpoints.ts
```diff
@@ -0,0 +1,76 @@
...
+import { SigningAlgorithm } from "@/types/files"
+import { CombineFormValues } from "@/types/forms"
+
+const BaseShape = z.object({
+  url: z
+    .string({ required_error: "URL is required" })
+    .trim()
+    .min(1, "URL is required")
+    .url("Must be a valid URL")
+    .startsWith("https://", "URL must use the https protocol"),
```
ezekg:
```suggestion
    .startsWith("https://", "URL must use the HTTPS protocol"),
```

ezekg (dismissed):
One minor capitalization issue, but other than that, looks good!

ezekg (approved):
LGTRM

Issue #229, opened 2026-06-16
Add account registration


ezekg:
No plan selection right now, just default to the free Dev 1 tier, plan ID `9e2d921a-96d7-44d6-82d5-88d25f6e7f65`.

Issue #226, opened 2026-06-16
Add full multiplayer mode support for Cloud


ezekg:
To be able to launch the Portal beta for Cloud, we'll need the sub-issues tackled.

PR #233, opened 2026-06-17
Refactor auth flow to standardize with established patterns

### LINKED ISSUE #227: Remove requirement for VITE_KEYGEN_ACCOUNT_ID env var in multiplayer mode
This shouldn't be required in multiplayer mode.

### LINKED ISSUE #228: Add account field before login when in multiplayer mode
Add an account selection field before login and other things like password reset requests. I think storing accounts into local storage might be a nice UX, so we can have a "last used" account that's automatically selected, e.g.:

Image

Should allow a text field input as well to type account slug or ID like the current dashboard.

Cazden:
Closes #227, closes #228. Related to #226 and #229.

Refactors the auth flow to have the same component API as other resources do, i.e. `/components/auth/form/account.tsx`, along with relevant schemas and making use of our established `<Forms />` component API.

Reason for this refactor is this allows us to properly expand the auth flow for cases like #228 and #229 without having to double back and refactor its implementation later, because it wasn't up to date with the rest of the codebase. We'll be able to implement both in a single pass.

---
If no account ID is available via environment, auth flow will now prompt for account or slug:

Then proceeds to standard login flow, with a helpful `Use a different account` button to return and change, if needed:

Also includes a revised register form, pending API integration in another PR:


ezekg:
fwiw, I think I'd prefer to keep the account step separate, so don't consolidate yet.

src/components/auth/form/fields.tsx
```diff
@@ -0,0 +1,303 @@
...
+}
+
+function AccountField({ autoFocus, fieldVariant }: FieldProps) {
+  const form = useFormContext<{ account: string }>()
+
+  const options = useMemo(
+    () =>
+      getRecentAccounts().map((account) => ({
+        value: account.id,
+        label: account.name ?? account.id,
```
ezekg: I'd prefer this to use account's name and the first 8 chars of the ID on the right of the dropdown option, like in the command bar's search results, and in the expiration heatmap license list.

ezekg: Also remember to add a "last used" badge next to the name if the account is the last used account (i.e. top of recent accounts).

src/components/auth/form/account.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const form = useForm<Schemas.Auth.AccountValues>({
+    resolver: zodResolver(Schemas.Auth.AccountSchema),
+    mode: "onChange",
+    defaultValues: { account: "" },
+  })
+
+  function onSubmit({ account }: Schemas.Auth.AccountValues) {
+    const id = account.trim()
+
+    addRecentAccount({ id })
```
ezekg: We should only add a recent account after successful login so that we can always store `id` and `name` (and storing invalid or inaccessible accounts would be a poor UX).

src/components/auth/form/register.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const form = useForm<Schemas.Auth.RegisterValues>({
+    resolver: zodResolver(Schemas.Auth.RegisterSchema),
+    mode: "onChange",
+    defaultValues: { username: "", password: "", companyId: "" },
+  })
+
+  // TODO(cazden) API call to check if email is available
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async function checkEmailAvailability(_email: string) {
```
ezekg: There is no such API endpoint. Just handle the errors if any, but it's impossible for a collision to happen here since emails are unique per-account, so a new account will never throw an `EMAIL_TAKEN` error.

src/components/auth/form/register.tsx
```diff
@@ -0,0 +1,137 @@
...
+import * as Loading from "@/components/loading"
+
+export default function RegisterForm() {
+  const navigate = useNavigate()
+  const [showIdField, setShowIdField] = useState(false)
+
+  const form = useForm<Schemas.Auth.RegisterValues>({
+    resolver: zodResolver(Schemas.Auth.RegisterSchema),
+    mode: "onChange",
+    defaultValues: { username: "", password: "", companyId: "" },
```
ezekg: I'm assuming this is from a slop-machine, but we don't have `username` or `companyId` concepts. 😅

Let's make sure to follow the API's naming, i.e. `user.email` and `account.slug`, respectively.

src/components/auth/form/register.tsx
```diff
@@ -0,0 +1,137 @@
...
+        type: "manual",
+        message: "This email is already registered.",
+      })
+      return
+    }
+
+    // Extract domain and slug from email
+    const parts = values.username.split("@")
+    const segments = parts[1].split(".")
+    const domain = segments.slice(0, 2).join("-").toLowerCase()
```
ezekg: The API already does this, so you can just leave `slug` blank during initial registration and only show the field if there was an error, like a `SLUG_TAKEN` collision -- i.e. somebody with `example.com` already registered, taking the `example-com` slug. That's what the current dashboard does iirc.

src/components/auth/form/register.tsx
Cazden: These were pre-form refactor with `/schemas` and the like, and iirc also pre-tanstack query usage. I'd say they're irrelevant to the current PR and are going to be addressed in a separate PR for #229. I'd disregard both this and `src/components/auth/form/recovery.tsx` for this PR, neither are really the focus because both forms are non-functional, and the only reason they're listed as additions here is because they were moved from their page component locations to form components, so purely housekeeping.

Embarrassing to see my old code. 👀

Cazden: Gonna mark these as closed if that's fine, since I'm planning to address the register flow in a different PR.

src/components/auth/form/register.tsx
Cazden: Marking this as resolved too. See https://github.com/keygen-sh/keygen-portal/pull/233#discussion_r3437303479.

src/components/auth/form/register.tsx
Cazden: Marking this as resolved too. See https://github.com/keygen-sh/keygen-portal/pull/233#discussion_r3437303479.

src/components/auth/form/login.tsx
```diff
@@ -0,0 +1,556 @@
...
+                  className="text-content-main underline-slide font-bold"
+                >
+                  Use a different account
+                </Link>
+              </Button>
+            </div>
+          )}
+
+          <div className="space-x-2 rounded border border-content-subdued p-2 text-center text-sm select-none">
+            <span className="text-content-subdued">No account yet?</span>
```
ezekg: This should be on the account step, not after they've already inputted an account.

ezekg: Also, change verbiage "Don't have an account? Create one" to match the callout on the registration.

src/components/auth/form/login.tsx
```diff
@@ -0,0 +1,556 @@
...
+                asChild
+                variant="link"
+                size="link"
+                className="text-content-loud"
+              >
+                <Link
+                  to="/auth"
+                  className="text-content-main underline-slide font-bold"
+                >
+                  Use a different account
```
ezekg: Move this above the fields, under the "Sign in to your account" heading.

src/components/suggest-input.tsx
```diff
@@ -0,0 +1,174 @@
...
+                <button
+                  key={option.value}
+                  type="button"
+                  role="option"
+                  aria-selected={index === activeIndex}
+                  onMouseEnter={() => setActiveIndex(index)}
+                  onMouseDown={(e) => e.preventDefault()}
+                  onClick={() => select(option.value)}
+                  className={cn(
+                    "flex w-full cursor-pointer items-center gap-2 rounded-sm px-2 py-1.5 text-left text-sm",
```
ezekg, line 145: Bottom padding for the dropdown seems a bit off:

Image

Should be the same for all sides.

src/keygen/config.ts
```diff
@@ -9,7 +16,14 @@ const config = {
     import.meta.env.VITE_KEYGEN_MODE === "multiplayer" &&
     CLOUD_HOSTS.includes(import.meta.env.VITE_KEYGEN_HOST),
   version: import.meta.env.VITE_KEYGEN_VERSION,
-  id: import.meta.env.VITE_KEYGEN_ACCOUNT_ID,
+
+  get id(): string {
+    return import.meta.env.VITE_KEYGEN_ACCOUNT_ID || accountFromUrl()
```
ezekg: This `accountFromUrl` fallback seems very brittle and requires that we keep `RESERVED_SEGMENTS` up-to-date.

Is this really needed vs keeping actual state elsewhere?

src/components/auth/form/login.tsx
```diff
@@ -0,0 +1,556 @@
...
+    keygen.client.setTokenId(tokenId)
+
+    if (!keygen.config.isCloud) {
+      storage.setItem("token", token)
+      keygen.client.setRootToken(token)
+    }
+
+    session.setUser(userId)
+
+    void navigate({ to: "/" })
```
ezekg: We should redirect to the account ID. Right now if you login using the slug, it keeps the slug in the URL after login, but that's brittle and could break if the slug is changed.

We always need to redirect to the ID once we have it.

src/components/auth/form/register.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  <Forms.Section.Header variant="auth">
+                    Create an account
+                  </Forms.Section.Header>
+                  <h2 className="text-sm text-content-subdued">
+                    You're one step away from joining Keygen.
+                  </h2>
+                </div>
+
+                <Auth.Form.Fields
+                  include={["email", "password"]}
```
ezekg: The margins between the fields feels a bit much, making them feel disconnected:

Image

I'd adjust the margins like this:

Image

ezekg (changes requested):
Looking good. A few minor changes.

ezekg (approved):
Looks good. I pushed a couple minor design tweaks, mainly for the email step's "current account" and I also added "current email" to the password step, with tooltips:

Image

Image

Issue #234, opened 2026-06-19
Love your project and wanted to contribute my Dockerfile if you need it


r4id4h:
Hey,

I was just checking out the project and since you were promoting it as nearly finished I took one of my Dockerfiles and created a beta build 😄 . It looks really nice, keep it up I really like it and will already start using it. If you don't have any docker builds yet I can contribute a PR.

best regards

ezekg:
Appreciate the kind words! We'll be adding a Dockerfile soon if we deem it a requirement for self-hosting, but ideally most people will self-host Portal as a static build deployed into an S3 bucket. At first glance, Docker is probably a bit overkill for most, but we'll get there when we get there. We'll know more once we start hosting Portal ourselves.

PR #235, opened 2026-06-19
Add dynamic env based configuration and docker build

r4id4h:
Linked to #234 

### Runtime configuration for Vite builds
Vite inlines import.meta.env.VITE_* values at build time, which means a single built artifact can't be reconfigured without rebuilding. This PR introduces runtime configuration so the same Docker image can be deployed with different settings via environment variables — relevant for CE/EE self-hosting where users configure their own deployment.

### How it works
A config.js accessor reads from a window.__KEYGEN_CONFIG__ object at runtime, falling back to import.meta.env for local dev.
On container start, an entrypoint script generates config.js from the container's environment variables (with sensible CE defaults).
App code keeps using the same config object — no call sites change beyond the central config.ts.

### Changes

- Add src/lib/env.ts accessor (runtime config → import.meta.env fallback)
- Update config.ts to read via the accessor
- Add public/config.js default and load it in index.html
- Add docker/entrypoint.sh to generate config.js from env vars at startup


ezekg:
Thanks for contributing, but we're going to be handling this aspect of self-hosting ourselves.

PR #237, opened 2026-06-23
Add account registration API integration

### LINKED ISSUE #226: Add full multiplayer mode support for Cloud

### LINKED ISSUE #229: Add account registration

Cazden:
Closes #229, closes #226. Picking up from the UI work completed in #233, adds account registration API integration to Portal.

Toasts/points to fields upon failed registration with error details:

And successfully registering automatically authenticates the user:


src/components/auth/form/register.tsx
```diff
@@ -23,11 +32,47 @@ export default function RegisterForm() {
...
+      localStorage.setItem("tokenId", data.id)
+      if (!keygen.config.isCloud) {
+        localStorage.setItem("token", data.attributes.token)
+      }
+
+      // *keygen music intensifies*
+      setIsRegistered(true)
+      await new Promise((resolve) => setTimeout(resolve, REDIRECT_DELAY_MS))
+
+      window.location.href = `/${account.attributes.slug}/app/dashboard`
```
ezekg: We should be using `navigate()` here. I'd assume this is a shortcut to reset state, but let's do it properly and hit the same code paths as login via https://github.com/keygen-sh/keygen-portal/pull/237/changes#r3467412232, i.e. make register do what login already does: persist token, update `keygen.client`, call `session.setUser()`, etc. then `navigate()` to the main dashboard.

src/components/auth/form/register.tsx
```diff
@@ -23,11 +32,47 @@ export default function RegisterForm() {
...
+          },
+        )
+      }
+
+      sessionStorage.removeItem("tokenId")
+      sessionStorage.removeItem("token")
+      localStorage.setItem("tokenId", data.id)
+      if (!keygen.config.isCloud) {
+        localStorage.setItem("token", data.attributes.token)
+      }
```
ezekg: Seems like this should be extracted into the authentication state. At the very least, leave a comment here explaining why this is here, e.g. to always persist the session as "remember me."

Personally, I'd wrap all of this stuff into a `login` utility, with a `logout` counterpart, then call that from the register and login routes.

src/components/auth/form/register.tsx
```diff
@@ -23,11 +32,47 @@ export default function RegisterForm() {
...
+
+      // *keygen music intensifies*
+      setIsRegistered(true)
+      await new Promise((resolve) => setTimeout(resolve, REDIRECT_DELAY_MS))
+
+      window.location.href = `/${account.attributes.slug}/app/dashboard`
+    } catch (error) {
```
ezekg, line 73: This error was being swallowed for me, I had to add `console.error` to realize the issue was a missing `VITE_KEYGEN_DEFAULT_PLAN_ID ` env var. We need to make sure errors are properly logged so that they make it into BetterStack, especially critical errors like these in the registration form.

ezekg (changes requested):

src/keygen/login.ts
```diff
@@ -0,0 +1,35 @@
...
+  storage.setItem("tokenId", tokenId)
+  client.setTokenId(tokenId)
+
+  // self-hosted has no cookie, so raw token is persisted and sent as a bearer.
+  if (!config.isCloud) {
+    storage.setItem("token", token)
+    client.setRootToken(token)
+  }
+
+  client.setAccount(accountId)
```
ezekg, line 32: I think we need to `setUser` here too, no?

ezekg (approved):

PR #242, opened 2026-06-25
Update older pages to align with newer patterns

### LINKED ISSUE #200: Refactor dated resource pages with newer components
As our component API has grown, need to go back and refactor some earlier developed resource pages to bring them up-to-date with the rest of the app, which includes:

* Products
* Entitlements

Some examples being non-hoverable tags/badges:

And attributes using just `<Attribute.Field />`...

...versus using `<Attribute.Field />` with `<Attribute.Value />` like most other pages do:

Cazden:
Closes #200.

Environments, products and entitlements now have hoverable badges and attributes like other details pages:

Products and entitlements now have advanced dialogs like other details pages:

### DIFF
--- a/src/components/entitlements/advanced-dialog.tsx
+++ b/src/components/entitlements/advanced-dialog.tsx
@@ -1,0 +1,114 @@
+import { useState } from "react"
+
+import { Skeleton } from "@/components/ui/skeleton"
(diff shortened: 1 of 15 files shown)


ezekg (approved):
Looks good, lgtm!

PR #254, opened 2026-06-30
Add API tokens UI and API integration

### LINKED ISSUE #248: Add API token management
This should have been included in the v1 milestone but was missed. Need to add API token management so users can integrate Keygen. Right now, the Access tab is empty. I'm still on the fence w.r.t. whether or not we even need Access, instead moving tokens to Settings > Developers. We could even have both, where admin/environment/product tokens are under Developers, and then have the tokens page lists license/user tokens by default, with the ability to clear the filter -- so pretty similar to what we're already doing for the users page.

Either way, here's the list of token types:

- [ ] Admin tokens
- [ ] Environment tokens
- [ ] Product tokens
- [ ] License tokens
- [ ] User tokens

And here's the sad sidebar:

Image

Cazden:
Closes #248. Adds API tokens support to Portal, accessible across two pages: `Access > Tokens` for managing tokens that are pre-filtered for `user` and `license`, typically for managing customer/user-related tokens, and then a separate section in `Settings > Developers` for managing `admin`, `environment` and `product`, and other internal role tokens, which are typically used for integrations.

### DIFF
--- a/src/components/sidebar/panel.tsx
+++ b/src/components/sidebar/panel.tsx
@@ -235,7 +235,14 @@ const VIEWS: View[] = [
     id: ViewId.Access,
     label: "Access",
     icon: KeyRound,
-    routes: [],
+    routes: linkOptions([
+      {
+        to: "/$accountId/app/tokens",
+        label: "Tokens",
+        params: { accountId: keygen.config.id },
+        requires: ["token.read"],
+      },
+    ]),
   },
   {
     id: ViewId.Security,
--- a/src/components/sidebar/panel.tsx
+++ b/src/components/sidebar/panel.tsx
@@ -46,7 +46,6 @@ import {
   Webhook,
   Package,
   Settings,
-  KeyRound,
 } from "lucide-react"
 
 import * as keygen from "@/keygen"
@@ -232,9 +231,9 @@ const VIEWS: View[] = [
     ]),
   },
   {
-    id: ViewId.Access,
-    label: "Access",
-    icon: KeyRound,
+    id: ViewId.Security,
+    label: "Security",
+    icon: Shield,
     routes: linkOptions([
       {
         to: "/$accountId/app/tokens",
@@ -242,13 +241,6 @@ const VIEWS: View[] = [
         params: { accountId: keygen.config.id },
         requires: ["token.read"],
       },
-    ]),
-  },
-  {
-    id: ViewId.Security,
-    label: "Security",
-    icon: Shield,
-    routes: linkOptions([
       {
         to: "/$accountId/app/event-logs",
         label: "Event Logs",
--- a/src/components/tokens/filter-bar.tsx
+++ b/src/components/tokens/filter-bar.tsx
@@ -1,0 +1,59 @@
+import { KeyRound } from "lucide-react"
+
+import * as Filters from "@/components/filter-bar"
+
+import config from "@/keygen/config"
+
+import { TokenBearerType, type TokenFilters } from "@/types/tokens"
+
+const BEARER_TYPES: ReadonlyArray<{
+  value: TokenBearerType
+  label: string
+  resource?: "users" | "licenses" | "products"
+}> = [
+  { value: TokenBearerType.User, label: "User", resource: "users" },
+  { value: TokenBearerType.License, label: "License", resource: "licenses" },
+  { value: TokenBearerType.Product, label: "Product", resource: "products" },
+  ...(config.isCE
+    ? []
+    : [{ value: TokenBearerType.Environment, label: "Environment" }]),
+]
+
+interface TokenFilterBarProps {
+  filters: TokenFilters
+  onChange: (filters: TokenFilters) => void
+}
+
+export default function TokenFilterBar({
+  filters,
+  onChange,
+}: TokenFilterBarProps) {
+  const filterCount = filters.bearerId ? 1 : 0
+
+  const value =
+    filters.bearerType && filters.bearerId
+      ? { type: filters.bearerType, id: filters.bearerId }
+      : undefined
+
+  return (
+    <Filters.FilterBar
+      filterCount={filterCount}
+      onClearAll={() => onChange({})}
+    >
+      <Filters.PolymorphicResourceFilter
+        label="Bearer"
+        icon={KeyRound}
+        placeholder="Bearer ID"
+        types={BEARER_TYPES}
+        value={value}
+        onChange={(next) =>
+          onChange(
+            next
+              ? { bearerType: next.type as TokenBearerType, bearerId: next.id }
+              : {},
+          )
+        }
+      />
+    </Filters.FilterBar>
+  )
+}
--- a/src/components/tokens/filter-bar.tsx
+++ b/src/components/tokens/filter-bar.tsx
@@ -2,8 +2,6 @@ import { KeyRound, Shield } from "lucide-react"
 
 import * as Filters from "@/components/filter-bar"
 
-import config from "@/keygen/config"
-
 import {
   TokenBearerType,
   TokenRole,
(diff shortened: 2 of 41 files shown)


src/components/tokens/filter-bar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  value: TokenBearerType
+  label: string
+  resource?: "users" | "licenses" | "products"
+}> = [
+  { value: TokenBearerType.User, label: "User", resource: "users" },
+  { value: TokenBearerType.License, label: "License", resource: "licenses" },
+  { value: TokenBearerType.Product, label: "Product", resource: "products" },
+  ...(config.isCE
+    ? []
+    : [{ value: TokenBearerType.Environment, label: "Environment" }]),
```
ezekg: Missing `resource` here, otherwise resource search bar is going to be broken, right?

src/keygen/tokens/create-for-bearer.ts
```diff
@@ -0,0 +1,54 @@
...
+  environmentToken,
+}: CreateForBearerProps): Promise<TokenResponse> {
+  const isLicense = values.bearerKind === TokenBearerKind.License
+
+  const body = {
+    data: {
+      type: "tokens",
+      attributes: compact({
+        name: values.name || undefined,
+        expiry: isLicense ? values.expiry : undefined,
```
ezekg: Expiry should not be limited to only license tokens, sometimes users like to add an expiry to their integration tokens forcing them to rotate them e.g. yearly.

src/keygen/tokens/create.ts
```diff
@@ -1,15 +1,45 @@
...
+  const isLicense = values?.bearerKind === TokenBearerKind.License
+
+  const body = {
+    data: {
+      type: "tokens",
+      ...(values != null
+        ? {
+            attributes: compact({
+              name: values.name || undefined,
+              expiry: isLicense ? values.expiry : undefined,
```
ezekg: See https://github.com/keygen-sh/keygen-portal/pull/254/changes#r3500356916

src/keygen/tokens/regenerate.ts
```diff
@@ -0,0 +1,20 @@
...
+interface RegenerateProps {
+  id: string
+}
+
+export default async function regenerate({
+  id,
+}: RegenerateProps): Promise<TokenResponse> {
+  const result = (await client.request(`/accounts/${config.id}/tokens/${id}`, {
+    method: "PUT",
+    root: true,
```
ezekg: Tokens are not always root, this will fail for a token within an environment.

src/lib/tokens.ts
```diff
@@ -0,0 +1,37 @@
...
+  kind: "raw",
+  name: "raw",
+  expiry: "date",
+  activations: "number",
+  maxActivations: "number",
+  deactivations: "number",
+  maxDeactivations: "number",
+}
+
+const BASE =
```
ezekg: Nit: name seems very unclear, prefer something like `BASE_REVOCATION_DESCRIPTION`.

src/lib/tokens.ts
```diff
@@ -0,0 +1,37 @@
...
+  token: Token | null | undefined,
+): string {
+  if (!token) return BASE
+
+  if (token.id === keygen.client.currentTokenId) {
+    return `${BASE} You're currently authenticated with this token, so revoking it will sign you out and require you to re-authenticate.`
+  }
+
+  if (token.attributes.kind === "admin-token") {
+    return `${BASE} Admin tokens can authenticate your portal session, so revoking this token may sign you out and require you to re-authenticate.`
```
ezekg: Nit: "Portal" should be capitalized.

src/queries/tokens.ts
```diff
@@ -1,49 +1,190 @@
...
-      }),
-    mutateAsync: ({ environmentId }: { environmentId: string }) =>
-      base.mutateAsync({
+    ...query,
+    data: query.data?.data ?? [],
+    links: query.data?.links,
+  }
+}
+
+export function useEnsureEnvironmentToken() {
```
ezekg: Nit: a better name might be `useCachedEnvironmentToken`, or even something like `useGetOrCreateEnvironmenToken` might be more conventional for these types of find-or-create operations.

src/components/tokens/form/fields.tsx
```diff
@@ -0,0 +1,590 @@
...
+  autoFocus,
+}: TokensFormFieldsProps) {
+  const fields = include ?? DEFAULT_FIELDS
+
+  return (
+    <>
+      {fields.map((field) => {
+        switch (field) {
+          case "bearerKind":
+            return <BearerKindField key="bearerKind" />
```
ezekg: Nit: should be called `BearerTypeField` so we don't confuse with the readonly `kind` attribute.

src/components/tokens/form/fields.tsx
```diff
@@ -0,0 +1,590 @@
...
+          case "bearerId":
+            return <BearerField key="bearerId" fieldVariant={fieldVariant} />
```
ezekg: Nit: should be called `BearerIdField` so we don't confuse it as a polymorphic field.

ezekg (commented):
First pass, mostly nits so feel free to ignore where you see fit. Will wait to do the full review and testing until after the new filters are added.

src/components/tokens/filter-bar.tsx
Cazden: Ended up removing this altogether because not only is the shape different from the others, but filtering by environment ID is kind of redundant when these tokens would only be accessed from their respective environments. Let me know if I'm missing something though.

src/hooks/use-token-table-columns.tsx
```diff
@@ -0,0 +1,65 @@
...
+      }),
+      column.attr("name", {
+        header: "Name",
+        cell: (info) =>
+          info.getValue() || <span className="text-content-muted">--</span>,
+      }),
+      column.rel("bearer", {
+        header: "Bearer",
+        enableSorting: false,
+        cell: (info) => <ResourceLink linkage={info.getValue()?.data} />,
```
ezekg: This needs to truncate long bearer identifiers:

Image

src/components/sidebar/panel.tsx
```diff
@@ -235,7 +235,14 @@ const VIEWS: View[] = [
     id: ViewId.Access,
     label: "Access",
     icon: KeyRound,
```
ezekg: Key icons should only be used for licenses/licensing, to reduce any confusion.

Honestly, I say we just roll this up into Security.

src/components/tokens/filter-bar.tsx
```diff
@@ -0,0 +1,81 @@
...
+            ...filters,
+            bearerType: next?.type as TokenBearerType | undefined,
+            bearerId: next?.id,
+          })
+        }
+      />
+      <Filters.ArrayFilter
+        label="Role"
+        icon={Shield}
+        options={ROLE_OPTIONS}
```
ezekg, line 72: There's some weird height issue on the filter:

Image

src/components/tokens/filter-bar.tsx
```diff
@@ -0,0 +1,81 @@
...
+} from "@/types/tokens"
+
+const BEARER_TYPES: ReadonlyArray<{
+  value: TokenBearerType
+  label: string
+  resource: "users" | "licenses" | "products"
+}> = [
+  { value: TokenBearerType.User, label: "User", resource: "users" },
+  { value: TokenBearerType.License, label: "License", resource: "licenses" },
+  { value: TokenBearerType.Product, label: "Product", resource: "products" },
```
ezekg, line 20: If we're in an environment, we should be able to filter to that environment's tokens:

Image

I know you can use role, but it's not obvious why environment is missing from the bearer filter, so it will likely be missed by users and cause needless support emails.

ezekg (changes requested):
Couple minor changes, but looks good.

src/components/tokens/view/internal.tsx
```diff
@@ -0,0 +1,134 @@
...
+          <Button
+            variant="outline"
+            size="sm"
+            onClick={() => setCreateOpen(true)}
+            className="border-none bg-background-2"
+          >
+            New Token
+          </Button>
+        </div>
+      </Can>
```
ezekg, line 52: Padding seems off vs the other panels:

Image

src/components/tokens/filter-bar.tsx
ezekg: Also, some of my comments were lost somehow, gg GitHub, but also notice that the current user shows as `ERROR` so something's off with the env-scoping for the current user here.

ezekg (approved):
Looks good! LGTM!

PR #270, opened 2026-07-06
Refactor remember me to be true by default

### LINKED ISSUE #265: Update login to use "remember me" by default
I'd guess 99% of users want to keep their session persistent and not have to login again later, so let's check this by default.

Cazden:
Closes #265.


ezekg (approved):
Image

PR #276, opened 2026-07-07
Add reset password API integration

Cazden:
Adds API support to the forgot password authentication flow, with some updated UI.


src/queries/users.ts
```diff
@@ -301,6 +301,26 @@ export function useResetPassword() {
...
 
+export function useResetPassword() {
+  return useMutation<
+    void,
+    APIError,
+    { email: string; token: string; newPassword: string }
+  >({
+    mutationFn: async ({ email, token, newPassword }) => {
+      const response = await keygen.users.resetPassword({
+        id: email,
```
ezekg: We don't need to ask the user for their email to fulfill a password reset, this is supplied [in the password reset token itself sent via email](https://github.com/keygen-sh/keygen-api/blob/2db5eb99d5b3fa360ab6b5ec36f81cce2f18d024/app/mailers/user_mailer.rb#L13), which should be split on `.` (e.g. [by the legacy dashboard](https://github.com/keygen-sh/keygen-app/blob/ec8481efc9a402f65377282e705776922497c8c9/app/routes/recover.js#L133)) to get all the required information to fulfill the password reset.

Per [the password reset mailer](https://github.com/keygen-sh/keygen-api/blob/2db5eb99d5b3fa360ab6b5ec36f81cce2f18d024/app/views/user_mailer/password_reset.html.erb#L8), the token will use the format `<account_id>.<user_id>.<random_token>`.

ezekg (changes requested):
Looks good, but email shouldn't be required for fulfillment. Clarified current behavior and token format.

ezekg (approved):
Looks good! lgtm!

PR #278, opened 2026-07-08
Add slug field to registration form

### LINKED ISSUE #273: Allow slug to be manually set on conflict during registration
Sometimes customers have legitimate reasons for "duplicate" accounts (those for the same email domain). When a user attempts to register using an email address domain that has already been registered, they receive an error from the API e.g. `SLUG_TAKEN`, which is currently unhandled. After a `slug` error, they should be able to manually set the account slug via a field.

Here's what the legacy registration form looks like after such an error, for reference:

Image

Cazden:
Closes #273. Adds a new slug field to the registration form that's only visible after trigging a `SLUG`-related error from the API after submitting, i.e. `SLUG_TAKEN`

Standard form before error:

After `SLUG_TAKEN` error:

### DIFF
--- a/src/components/auth/form/register.tsx
+++ b/src/components/auth/form/register.tsx
@@ -5,7 +5,7 @@ import { zodResolver } from "@hookform/resolvers/zod"
 
 import { Button } from "@/components/ui/button"
 
-import { cn, dasherize } from "@/lib/utils"
+import { cn } from "@/lib/utils"
 import { handleFormError } from "@/lib/form-errors"
 
 import * as Schemas from "@/schemas"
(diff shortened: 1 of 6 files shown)


src/components/auth/form/register.tsx
```diff
@@ -75,10 +76,20 @@ export default function RegisterForm() {
 
       console.error(error)
 
+      const apiError = error instanceof APIError ? error : undefined
+      const isSlugError =
+        apiError != null &&
+        (apiError.code?.startsWith("SLUG_") === true ||
```
ezekg: Nit: `=== true` is unnecessary since `startsWith` is already a boolean, and I'd argue that checking the error code for a prefix is also unnecessary. I'd prefer we just look at `source.pointer`.

ezekg (approved):
Looks good, and works good. But like I said in Discord, the "new" slug bit is starting to reveal some design issues with the current auth form fields. Feel free to address later in a separate PR, if you agree with the concerns.

PR #280, opened 2026-07-09
Add raw JSON support to metadata form fields

### LINKED ISSUE #272: Allow metadata field to accept raw JSON paste
Some users have said that they'd like to be able to toggle the metadata field to raw during resource edit/create and paste in a JSON object, instead of filling each metadata property individually, kind of like the individual JSON property type for metadata fields we already have, except for the entire `metadata` object.

Cazden:
Closes #272. Adds a toggle on all metadata fields for switching between `Fields` and `Raw` JSON input modes.

Fields is the default which has a key/value input:

Raw is an alternative input mode for metadata fields that allow users to write or paste valid JSON instead:

Raw includes some editor conveniences, such as tab indentation, indented newlines, auto-closing braces/quotes, etc. basically some general assists you'd expect from a code editor.

### DIFF
--- a/src/components/artifacts/form/fields.tsx
+++ b/src/components/artifacts/form/fields.tsx
@@ -28,7 +28,7 @@ import { type FieldVariant } from "@/components/forms/field"
 
 import * as Forms from "@/components/forms"
 import * as Search from "@/components/search"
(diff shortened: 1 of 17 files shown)


ezekg (approved):
Looks great! lgtm

PR #231, opened 2026-06-16
Add BetterStack error reporting

Cazden:
Closes #182.

Also adds a new `.env.sample` file with environment config instructions, mirroring what the API has.

PR #283, opened 2026-07-09
Add support for BetterStack source maps

Cazden:
Follow-up to  #231, adds readable error logging to our BetterStack config by adding support for source maps.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -5,6 +5,7 @@ import { execSync } from "node:child_process"
 import tailwindcss from "@tailwindcss/vite"
 import react from "@vitejs/plugin-react"
 import svgr from "vite-plugin-svgr"
+import { sentryVitePlugin } from "@sentry/vite-plugin"
 
 function resolveAppVersion(): string {
   try {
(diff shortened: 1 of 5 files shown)


vite.config.ts
```diff
@@ -37,6 +37,10 @@ function appVersion(): PluginOption {
 
 export default defineConfig(({ mode }) => {
   const env = loadEnv(mode, process.cwd(), "")
+  const version = resolveAppVersion()
+
+  // for source map upload to BetterStack
+  const sentryAuthToken = env.SENTRY_AUTH_TOKEN
```
ezekg, line 43: Should this be prefixed with `VITE_` or no? I don't know conventions lol but trying to set this up in Cloudflare.

Cazden: No it's intentionally left out: https://github.com/keygen-sh/keygen-portal/blob/214b5bff0d635f338ac63ac98198e3582e993f52/.env.sample#L51-L53

ezekg: lmfao would help if I could read huh

ezekg (approved):

PR #292, opened 2026-07-13
Refactor timestamps to display UTC

### LINKED ISSUE #123: Timestamps should be displayed in UTC
Right now, some timestamps are formatted into the user's timezone instead of UTC, which can result in inconsistent timestamps. This also creates confusing situations where a user is filtering on e.g. licenses expiring on the 30th, but nothing is shows up; yet if they were to filter on the 31st, results for the 30th show up:

Image

This is because the license showing up as the 30th has an expiry on `2026-03-31T00:00:00.000Z`, which is the 30th in CST timezone.

In addition, I think all timestamps should have a popover showing UTC and the user's local timezone:

Image

Cazden:
Closes #123. Changes timestamps across the codebase to be based on UTC, and adds a tooltip popover that shows a clipboard-able raw UTC timestamp, and formatted UTC/local timestamps (e.g. CST, EST).

In list pages:

In resources attributes:

In resources overview sidebar:

In resources event logs sidebar:

Timestamps shown as raw in event/request logs list:


src/components/event-logs/feed.tsx
```diff
@@ -79,9 +78,7 @@ function EventLogRow({
           </Badge>
         </div>
         <div className="mt-1 flex min-w-0 items-center gap-2 text-xs">
-          <span className="shrink-0 text-content-muted" title={timestamp}>
-            {formatRelativeTime(created)}
-          </span>
+          <Timestamp value={created} display="relative" className="shrink-0" />
           <span className="min-w-0 truncate text-content-normal">
```
Cazden, line 82: Redundant to pass in `relative` when we already default to that display value.

src/components/tokens/view/internal.tsx
```diff
@@ -129,9 +130,13 @@ function TokenRow({ token, onNavigate }: TokenRowProps) {
           )}
         </div>
         <span className="text-xs text-content-subdued">
-          {token.attributes.expiry
-            ? `Expires ${new Date(token.attributes.expiry).toLocaleDateString()}`
-            : "Never expires"}
+          {token.attributes.expiry ? (
+            <>
+              Expires <Timestamp value={token.attributes.expiry} />
+            </>
```
Cazden, line 139: This will read wrong if already expired, e.g. "Expires 3 days ago"

src/components/sidebar/billing-status-card.tsx
```diff
@@ -29,7 +28,7 @@ const STATE_CONTENT: Partial<Record<BillingState, StateContent>> = {
     title: "Your free trial ends soon",
     description: (end) =>
       end
-        ? `Your trial ends on ${formatDate(new Date(end), DATE_FORMAT)}. Upgrade to keep full access.`
+        ? `Your trial ends on ${formatUtcDate(end)}. Upgrade to keep full access.`
```
Cazden: Not sure this is correct. Since we're not showing any UTC/Local popover here, users not in UTC will potentially see an incorrect end date.

src/components/event-logs/feed.tsx
```diff
@@ -79,9 +78,7 @@ function EventLogRow({
           </Badge>
         </div>
         <div className="mt-1 flex min-w-0 items-center gap-2 text-xs">
-          <span className="shrink-0 text-content-muted" title={timestamp}>
-            {formatRelativeTime(created)}
```
Cazden, line 83: `shrink-0` is needed here to prevent wrapping to a newline and breaking styles.

ezekg (approved):
Looks really great! gg

PR #296, opened 2026-07-14
Add scheme to policy create forms

### LINKED ISSUE #291: Policy creation is missing the scheme attribute
https://keygen.sh/docs/api/policies/#policies-object-attrs-scheme

Should be shown when "offline" is checked during create.

### LINKED ISSUE #295: Scroll is broken in multi-dialog forms
Removing `<DialogOverlay />` via our `disableOverlay` prop and substituting with our own overlay via `<Forms.Container.Overlay />` introduced an issue with Radix where we inadvertently broke mouse-wheel scrolling, because the default overlay locks scroll then whitelists the dialog's content, and since we're using a separate overlay that has no content of its own in multi-dialog forms (e.g. policy create flow), nothing ends up whitelisted and the mouse-wheel doesn't scroll. We can work around this by styling with `bg-transparent` rather than removing the overlay entirely so the whitelisting works.

Possibly related to https://github.com/radix-ui/primitives/issues/1159.

Cazden:
Closes #291, closes #295. Adds the `scheme` attribute to policy template, scratch and duplicate forms. Also adds a new reusable `<Notice />` component for notice callouts when we need them, e.g. the admin token call out in `/app/tokens`.

Also fixes an issue (#295) where scroll via mouse wheel was broken in multi-dialog forms, such as the policy create forms here.

In the template flow, `scheme` is included in advanced with a collapsed card by default to keep it out of the way from normal user interaction:

Scratch form's `scheme` attribute is located in a collapsed card as well, at the bottom of the `Requirements` section:

Duplicate policy is located at the bottom of the form in a collapsible card that's closed by default, as well:

### DIFF
--- a/src/components/policies/form/duplicate.tsx
+++ b/src/components/policies/form/duplicate.tsx
@@ -22,6 +22,7 @@ import { settleCreateEntitlements } from "@/lib/entitlements"
 import * as Forms from "@/components/forms"
 import * as Loading from "@/components/loading"
 import * as Policies from "@/components/policies"
+import { Notice } from "@/components/notice"
 import { Separator } from "@/components/ui/separator"
 
 interface DuplicatePolicyFormProps {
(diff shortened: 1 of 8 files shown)


src/components/policies/form/duplicate.tsx
```diff
@@ -228,6 +229,30 @@ export default function DuplicatePolicyForm({
                 />
               </Forms.Section.Column>
             </Forms.Section.Columns>
+
+            <Separator className="my-8" />
+
+            <Forms.Section.Columns title="Offline configuration">
```
ezekg: Can we put this into a collapsible card too? I don't want a notice to be visible for every duplicate, as that's going to scare users about something that 99% don't care about or understand.

ezekg (approved):
LGTM!

PR #299, opened 2026-07-15
Fix broken breadcrumbs after certain navigations

### LINKED ISSUE #217: Breadcrumbs are broken after certain navigations
e.g. go to request logs > view detail > click the requestor user > click Users in breadcrumb, I'd expect to go to the Users list, but I go back to the request log detail. Attaching a video showcasing current behavior. I think this was introduced in #161.

Cazden:
Closes #217. Breadcrumbs used `useBackNavigate()` which used windows history, so it indiscriminately went to the last known path and would break under certain navigations, e.g. `/request-logs/$id` > requestor user page > breadcrumb would nav back to `/request-logs/$id` rather than the users list.

This fixes that with `useBreadcrumbBackNavigate()`, which conditionally handles navigating from list > details and preserves state (i.e. any applied filters) when navigating back to list, and navigating from e.g. relationship links and using the associated resource details page breadcrumbs to navigate to the list view works correctly.

Also addresses a Users/Teams coupling issue that was making navigations difficult to work with under certain conditions, since they were sharing the same details page component. It's now separated into `teams/` and `users/` with their own list and details pages.

### DIFF
--- a/src/components/resource-link.tsx
+++ b/src/components/resource-link.tsx
@@ -8,7 +8,6 @@ import {
 } from "@/components/ui/tooltip"
 
 import { type Linkage } from "@/types/api"
-import { InternalRoles } from "@/types/users"
 
 import { useGetArch } from "@/queries/arches"
 import { useGetUser } from "@/queries/users"
@@ -64,7 +63,6 @@ interface ResourceLinkViewProps {
   isError?: boolean
   buttonClassName?: string
   truncate?: boolean
-  to?: string
 }
 
 function ResourceLinkView({
@@ -74,7 +72,6 @@ function ResourceLinkView({
   isError,
   buttonClassName,
   truncate = false,
-  to = `/$accountId/app/${linkage.type}/$id`,
 }: ResourceLinkViewProps): ReactElement {
   if (isLoading) return <Skeleton className="h-5 w-32 rounded-sm" />
 
@@ -88,7 +85,7 @@ function ResourceLinkView({
       <Tooltip>
         <TooltipTrigger className="cursor-pointer">
           <GoToButton
-            path={to}
+            path={`/$accountId/app/${linkage.type}/$id`}
             params={{ accountId: keygen.config.id, id: linkage.id }}
             label={text}
             disabled
@@ -108,7 +105,7 @@ function ResourceLinkView({
 
   return (
     <GoToButton
-      path={to}
+      path={`/$accountId/app/${linkage.type}/$id`}
       params={{ accountId: keygen.config.id, id: linkage.id }}
       label={text}
       buttonClassName={buttonClassName}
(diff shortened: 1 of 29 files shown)


ezekg:
Just FYI, I think this was originally done because going to Settings > Team > Teammate Detail, clicking "Team" in the breadcrumbs went back to the user list page vs going back to the team settings page. I'm not sure if that's super important or not, but if it could be fixed, that would probably improve UX in those special cases. To be honest, that issue probably just comes from mixing concerns by making "Users" and "Team" synonymous.

Cazden:
> Just FYI, I think this was originally done because going to Settings > Team > Teammate Detail, clicking "Team" in the breadcrumbs went back to the user list page vs going back to the team settings page. I'm not sure if that's super important or not, but if it could be fixed, that would probably improve UX in those special cases.

Thanks for the info. I believe this particular case has been addressed and manually tests fine:
https://github.com/keygen-sh/keygen-portal/blob/8faeb700851e05ad2cc2580990f13dec52bd03ea/src/pages/app/user/details.tsx#L117-L118
https://github.com/keygen-sh/keygen-portal/blob/8faeb700851e05ad2cc2580990f13dec52bd03ea/src/pages/app/user/details.tsx#L195-L205

ezekg (dismissed):

src/components/resource-link.tsx
```diff
@@ -171,7 +174,29 @@ const RESOURCE_LINKS: Record<
...
+    buttonClassName,
+    truncate,
+  }: ResolvedLinkProps): ReactElement {
+    const { data, isLoading, isError } = useGetUser(linkage.id)
+
+    return (
+      <ResourceLinkView
+        linkage={linkage}
+        to={
+          data && InternalRoles.includes(data.attributes.role)
```
ezekg: This feels like we're right back to where we started with mixing concerns between Team and Users just because they both concern a user object. I'd say let's add `teams` and stop overloading the `users` key, but in this case, I don't see why this is even needed -- e.g. clicking a license owner that happens to be an admin should just view the user's details, regardless of their role. It doesn't need to link to the team page at all.

src/pages/app/team/details.tsx
```diff
@@ -0,0 +1,744 @@
...
+                          path="/$accountId/app/licenses"
+                          params={{ accountId: keygen.config.id }}
+                          search={{ user: id }}
+                          label="View all"
+                        />
+                      }
+                    />
+                    <Attribute.Field
+                      variant="text"
+                      label="Machines"
```
ezekg: Nit: not very relevant for teammates, so I vote to remove.

src/pages/app/team/details.tsx
```diff
@@ -0,0 +1,744 @@
...
+                      label="Group"
+                      value={
+                        <ResourceLink
+                          linkage={user.relationships.group?.data}
+                        />
+                      }
+                    />
+                    <Attribute.Field
+                      variant="text"
+                      label="Licenses"
```
ezekg: Nit: not very relevant for teammates, so I vote to remove.

src/pages/app/team/details.tsx
```diff
@@ -0,0 +1,744 @@
...
+                            tooltip={UserAttributeDescriptions.role}
+                          />
+                        }
+                      />
+                    </div>
+                  </div>
+                </CollapsibleCard>
+
+                <CollapsibleCard
+                  title="Products"
```
ezekg: Nit: not very relevant for teammates, so I vote to remove.

src/pages/app/team/details.tsx
```diff
@@ -0,0 +1,744 @@
...
+                  ) : (
+                    <Can permission="user.ban">
+                      <DropdownMenuItem
+                        onClick={(e) => {
+                          toggleOpen("ban", true)
+                          e.currentTarget.blur()
+                        }}
+                        className="pb-2 text-base"
+                      >
+                        Ban
```
ezekg: Nit: not very relevant for teammates, so I vote to remove.

ezekg (approved):

ezekg:
Looks great! LGTM!

PR #305, opened 2026-07-16
Add license permission fields

### LINKED ISSUE #142: Add ability to add custom permissions to users and licenses
Their permissions should be an editable field, set to the account's defaults when omitted, like for products.

Cazden:
Closes #142. Addresses a few issues with permissions and adds permissions config to licenses.

Licenses now includes a permissions form field, and permission sections have been added to the License, User and Team details pages. Additionally, this fixes a problem in CE that allowed users to attempt to modify and make API requests with permission attributes, which is an EE-exclusive feature, so the API would reject.

All forms that include permissions (products, licenses, users, invite teammate, and tokens) now have their permissions form field separated into its own step that requires EE to be viewable. Eventually we'll add a CTA so the steps are still viewable in CE (#306).

NB: moving the permission form fields into separate steps was necessary because step validation is performed per-step submit, so if we conditionally included the form field in a step with other fields, that would dirty validation because the step would be expecting permissions and wouldn't have access to the field in CE. Isn't really an issue today because permissions are always optional, but could be an issue if we ever required permissions in the same step with other fields.

### DIFF
--- a/src/components/collapsible-card.tsx
+++ b/src/components/collapsible-card.tsx
@@ -1,4 +1,4 @@
-import { useState, Fragment, Children } from "react"
+import { useState, Fragment, Children, isValidElement } from "react"
 import { motion, AnimatePresence } from "motion/react"
 
 import { cn } from "@/lib/utils"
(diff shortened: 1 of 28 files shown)


ezekg (changes requested):
Permissions are only available in EE, so this is broken in CE. Need to make sure permissions are gated to EE.

Image

ezekg (approved):
Looks good!

PR #307, opened 2026-07-17
Fix sidebar desync from certain navigations

### LINKED ISSUE #38: Sidebar category should always be in sync with the route
e.g. if we're on `Products`, the active sidebar section should be `Licensing`. Can desync whenever the user does something like refresh the page.

Cazden:
Closes #38.

### DIFF
--- a/src/components/sidebar/panel.tsx
+++ b/src/components/sidebar/panel.tsx
@@ -341,9 +341,17 @@ function useVisibleViews(): View[] {
 export default function SidebarPanel(): React.ReactElement {
   const activeView = useActiveView()
   const visibleViews = useVisibleViews()
+  const [selectedView, setSelectedView] = useState(activeView)
+  const [prevActiveId, setPrevActiveId] = useState(activeView.id)
+
+  // follow route when nav jumps into a different view section
+  if (activeView.id !== prevActiveId) {
+    setPrevActiveId(activeView.id)
+    setSelectedView(activeView)
+  }
+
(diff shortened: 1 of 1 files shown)


src/components/sidebar/panel.tsx
```diff
@@ -341,9 +341,17 @@ function useVisibleViews(): View[] {
 export default function SidebarPanel(): React.ReactElement {
   const activeView = useActiveView()
   const visibleViews = useVisibleViews()
+  const [selectedView, setSelectedView] = useState(activeView)
+  const [prevActiveId, setPrevActiveId] = useState(activeView.id)
+
+  // follow route when nav jumps into a different view section
+  if (activeView.id !== prevActiveId) {
+    setPrevActiveId(activeView.id)
+    setSelectedView(activeView)
```
ezekg, line 350: My gut says that setting state inline in the render is a bad idea, but maybe this is fine? I bounced it through a few LLMs and they all say this is the best solution to the problem, but it seems unsafe at first glance.

ezekg (approved):
I have reservations about the implementation, namely about risk of introducing an infinite loop via inline set state, but the reservations are probably just due to a lack of deep knowledge around React best practices. Approving since it passed multiple LLM reviews, even if my gut is saying otherwise.

src/components/sidebar/panel.tsx
Cazden: That was my first instinct as well, but also got the same response from LLMs.

Issue #288, opened 2026-07-13
Leaderboard: add ability to filter data from metrics, add more categories and aggregations


maxlund:
Because my integration tests run against actual keygen API, I have metrics like this.

<img width="1330" height="242" alt="Image" src="https://github.com/user-attachments/assets/8b5917ff-725c-4397-b125-7d622cc6802c" />

Would be nice to have the ability to filter out certain values.

Also would be nice to see the user email for the top licenses, and also an ability to aggregate usage across all licenses for a single email.
Perhaps also aggregate usage across email domains (see most active customers, etc).

Add some simple generic analytics ingest endpoints where we can set our own keys and values to count/store to track user behavior, and you've got yourself a competitor to Mixpanel & co! 🤑

PR #309, opened 2026-07-17
Add additional license info to Leaderboards chart

### LINKED ISSUE #289: Top licenses chart should show more information about licenses
Right now we only show the ID, which isn't very helpful. In the popover, we should show additional information like license name, owner name, etc., similar to the heatmap, but with a bit more information.

Image

Cazden:
Closes #289, and related to #288. Adds license `status` and `owner` to license tooltips in the Leaderboards chart. Open to suggestions if we want to include more data here.

### DIFF
--- a/src/components/analytics/leaderboards.tsx
+++ b/src/components/analytics/leaderboards.tsx
@@ -11,6 +11,7 @@ import {
 } from "recharts"
 
 import CursorTooltip from "@/components/cursor-tooltip"
+import { Badge } from "@/components/ui/badge"
 import { Card, CardContent } from "@/components/ui/card"
 import {
   ChartConfig,
@@ -23,11 +24,19 @@ import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs"
 
 import { useCursorFollowTooltip } from "@/hooks/use-cursor-follow-tooltip"
 
+import { useGetUser } from "@/queries/users"
+import { useGetLicense } from "@/queries/licenses"
 import { useLeaderboard } from "@/queries/analytics"
 
-import { GREEN, LEADERBOARDS, useLazyVisibility } from "@/lib/analytics"
+import { cn } from "@/lib/utils"
 import { truncator } from "@/lib/truncate"
+import { getUserLabel } from "@/lib/users"
+import { GREEN, LEADERBOARDS, useLazyVisibility } from "@/lib/analytics"
+
+import { type Linkage } from "@/types/api"
 import { HTTPMethods, type HTTPMethod } from "@/types/http"
+import { LicenseStatusLabels, LicenseStatusVariants } from "@/types/licenses"
+
 import * as keygen from "@/keygen"
 import EmptyChart from "./empty-chart"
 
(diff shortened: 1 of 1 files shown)


src/components/analytics/leaderboards.tsx
```diff
@@ -201,26 +210,121 @@ function LeaderboardCard({
...
+  const { data: user, isLoading, isError } = useGetUser(linkage.id)
+
+  if (isLoading) {
+    return <Skeleton className="ml-auto h-4 w-24 rounded-sm" />
+  }
+
+  if (isError || !user) {
+    return (
+      <span className="font-mono text-[11px] text-content-subdued">
+        {linkage.id}
```
ezekg: This should be displayed how we display other IDs, i.e. truncated.

src/components/analytics/leaderboards.tsx
```diff
@@ -201,26 +210,121 @@ function LeaderboardCard({
...
+        <p className="font-medium text-content-muted">License unavailable</p>
+      ) : (
+        <p className="truncate font-medium text-content-muted">
+          {license?.attributes.name || (
+            <span className="text-content-disabled">(name not set)</span>
+          )}
+        </p>
+      )}
+      <p className="mt-0.5 truncate font-mono text-[11px] text-content-subdued">
+        {licenseId}
```
ezekg, line 260: Fine to leave this untrucated imo, to go along with my other comment.

ezekg (dismissed):
Nit: I think we should truncate the owner ID, but since it's an exception (e.g. the owner doesn't exist), it's not a big deal either way to leave it and you can go ahead and merge if you want.

src/components/analytics/leaderboards.tsx
```diff
@@ -41,6 +41,7 @@ import * as keygen from "@/keygen"
 import EmptyChart from "./empty-chart"
 
 const truncateMiddle = truncator("middle", { maxLength: 34 })
+const truncateId = truncator("middle", { maxLength: 12 })
```
ezekg: Should be max length of 8:

- https://github.com/keygen-sh/keygen-portal/blob/64c7b0ce4a3624de8cd0c287a0d3e2beeac1b862/src/components/palette/search.tsx#L28
- https://github.com/keygen-sh/keygen-portal/blob/64c7b0ce4a3624de8cd0c287a0d3e2beeac1b862/src/components/auth/form/fields.tsx#L26
- https://github.com/keygen-sh/keygen-portal/blob/64c7b0ce4a3624de8cd0c287a0d3e2beeac1b862/src/components/analytics/license-expiration-heatmap.tsx#L60

ezekg: Also should use `clip`, not `middle`.

ezekg (approved):

maxlund:
Rathee than just as a tool tip on hover, it would be great if a filter view/view option would be to see the license email as key. Perhaps with aggregation as an option.

ezekg:
@maxlund new aggregations would be brand new charts, since aggregations are done server-side, i.e. not something that's client-side or configurable. Will continue to think about introducing new aggregations that will be beneficial for everybody. I definitely see a Top Users chart being valuable.

maxlund:
> @maxlund new aggregations would be brand new charts, since aggregations are done server-side, i.e. not something that's client-side or configurable. Will continue to think about introducing new aggregations that will be beneficial for everybody. I definitely see a Top Users chart being valuable.

Awesome! Seeing the most active emails, and maybe aggregate by domain as well (several users from same company). Email + license as a group-by key would be useful for us also, since customers often buy a bulk of licenses at e.g. "operations@company.com". Would be interesting to see if specific people (license keys) in the company are using it more than others.

PR #311, opened 2026-07-20
Add better handling of not found errors

### LINKED ISSUE #175: Add better handling of not found errors
e.g. pages that don't exist, e.g. `/app/foo` or resources that don't exist `/app/licenses/invalid`.

Cazden:
Closes #175.


src/pages/error/not-found.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <h1 className="font-owners-wide text-xl">Error 404</h1>
+        </div>
+        <div className="flex flex-col gap-2 p-4">
+          <p className="text-content-muted">
+            The page or resource you were looking for could not be found.
+          </p>
+          <p className="text-xs text-content-subdued">
+            If this error is unexpected, and the problem persists, please{" "}
+            <a
+              href="mailto:support@keygen.sh"
```
ezekg: Lets move this to an optional environment variable so that we don't direct self-hosted Portals to our support email. Maybe adjust wording based on if it's Cloud or not?

src/keygen/config.ts
```diff
@@ -32,6 +32,10 @@ const config = {
     return import.meta.env.VITE_KEYGEN_DEFAULT_PLAN_ID || ""
   },
 
+  get supportEmail(): string {
+    return import.meta.env.VITE_SUPPORT_EMAIL || ""
```
ezekg: Should be `VITE_KEYGEN_SUPPORT_EMAIL`. We should always namespace our config from others.

ezekg (changes requested):
Small change, but other than that, looks good!

ezekg (approved):

PR #312, opened 2026-07-21
Add route-based document titles

### LINKED ISSUE #141: Page titles should match current route
Tab title should reflect the current route, e.g. "License xxx."

Cazden:
Closes #141.

### DIFF
--- a/src/pages/app/artifact/details.tsx
+++ b/src/pages/app/artifact/details.tsx
@@ -76,6 +76,7 @@ import ResourceLink from "@/components/resource-link"
 import ConfirmationModal from "@/components/confirmation-modal"
 import TooltipBadge from "@/components/tooltip-badge"
 import CollapsibleCard from "@/components/collapsible-card"
+import DocumentTitle from "@/components/document-title"
 
 const ArtifactStatusIcons: Record<ArtifactStatus, React.ReactNode> = {
   [ArtifactStatus.Waiting]: <CircleDashed className="size-3" />,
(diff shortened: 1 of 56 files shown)


src/pages/app/artifact/details.tsx
```diff
@@ -124,6 +125,9 @@ export default function ArtifactDetails() {
 
   return (
     <section className="flex h-screen w-full">
+      <DocumentTitle
+        title={`Artifact: ${artifact?.attributes.filename || id}`}
+      />
```
ezekg, line 130: So just to understand, this implementation relies on the browser hoisting the `<title>` element from the `<body>` into the `<head>`? I'd argue there's gotta be a better, more conventional way to accomplish this than relying on a browser quirk...

Maybe something like what Oxide does with crumbs?

- https://github.com/oxidecomputer/console/blob/main/app/hooks/use-crumbs.ts
- https://github.com/oxidecomputer/console/blob/main/app/pages/project/disks/DisksPage.tsx#L50

That's built on React Router's `handle`:

- https://reactrouter.com/how-to/using-handle

Then render that into the head's title, or use an effect that sets `document.title` directly (idk what's conventional here tbqh).

ezekg: Never mind, I'm retarded, forgive me: https://egghead.io/hoist-title-and-meta-tags-in-react-19~h6z5l

ezekg (approved):

PR #315, opened 2026-07-22
Fix numeric inputs accepting non-numeric values

### LINKED ISSUE #174: Numeric input fields should become invalid on non-numeric input
The non-numeric input is accepted by the form, but stripped in the actual update request.

Image

### LINKED ISSUE #314: Policy duration form field breaks under certain conditions
Emptying the field, then selecting `Unlimited` as the unit seems to break the input, as any subsequent changes don't work, i.e. presets don't apply, and entering any value and trying to continue will fail validation because the form value itself is technically empty.

This seems to be isolated to Policy forms. Field is detaching from RHF somehow.

Cazden:
Closes #174. Also exposed a bug when working with Policy `duration` where we could break the input and prevent the user from being able to validate and finish policy creation. This occurred in a couple other policy fields as well, such as `requireHeartbeat` and `checkInInterval`, which was caused by resetting the form when the field's value met certain conditions (which we were using to e.g. reset `requireHeartbeat`s related fields to their default values when it was `false`). Fixed to just reset the fields now instead of the form, so this also closes #314.

`asdasda` has no power here:

### DIFF
--- a/src/components/artifacts/form/fields.tsx
+++ b/src/components/artifacts/form/fields.tsx
@@ -28,6 +28,7 @@ import { type FieldVariant } from "@/components/forms/field"
 
 import * as Forms from "@/components/forms"
 import * as Search from "@/components/search"
+import NumberInput from "@/components/number-input"
 import MetadataInput from "@/components/metadata-input"
 
 type Descriptions = typeof ArtifactFormFieldDescriptions
@@ -307,10 +308,8 @@ function FilesizeField({
             optional
           >
             <FormControl>
-              <Input
+              <NumberInput
                 {...field}
-                value={field.value ?? ""}
-                type="number"
                 placeholder="e.g. 3097"
                 autoFocus={autoFocus}
                 autoComplete="off"
--- a/src/components/byte-size-input.tsx
+++ b/src/components/byte-size-input.tsx
@@ -1,6 +1,5 @@
 import { useState } from "react"
 
-import { Input } from "@/components/ui/input"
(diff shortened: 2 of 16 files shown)


ezekg (approved):
Just so we have a papertrail, as I mentioned in Discord, the `form.reset()` was originally added in https://github.com/keygen-sh/keygen-portal/pull/100 to fix a performance issue with large forms, but it seems like that might have been caused by something else that we fixed since then as I don't see the performance issue crop up again here in the reverting of `form.reset()`. So with that, lgtm!

Cazden:
Found out why it's working now and did some manual testing to confirm it.

```
field.onChange(value)
if (mode === PolicyMode.Create && value === null) {
  form.resetField("expirationStrategy")
  form.resetField("expirationBasis")
  form.resetField("renewalBasis")
  form.resetField("transferStrategy")
}
```

This *does* reintroduce the performance issue, but it won't happen in the edit flow. For it to occur, this condition has to pass, meaning besides gating to the create flow, it also only occurs on the final backspace, e.g. backspacing `123456789` would only impact performance on the final backspace of `1` when the value becomes `null` and `form.resetField()` fires.

So essentially, in #100, you fixed the issue twice over by moving the code into `PolicyMode.Create` AND using `form.reset()`, but `form.reset()` introduced the issue that I fixed related to the field desyncing from RHF, so *now* our current solution is to just rely on one fix via gating to `PolicyMode.Create`.

tl;dr #100 implemented two fixes for the issue, I just reverted one, and that's why it still works.

PR #323, opened 2026-07-28
Fix list state persistence

### LINKED ISSUE #320: Token list filters aren't persisted to the URL
When filtering tokens, e.g. by role or bearer, the filters aren't persisted to the URL query parameters like other filters. This causes the "go back" functionality to break, since filters are cleared upon returning to the list page.

### LINKED ISSUE #321: Pagination cursor is not persisted to URL
When paginating a resource list, the pagination cursor is not persisted to the URL, so clicking a row and then going back returns to the first page. The cursor should be persisted to the URL so that back navigating works correctly.

Cazden:
Closes #320, closes #321.

Paginating persists the cursor to the URL, so subsequently clicking into a resource and hitting back returns you to the page you were on instead of the first.

Also fixes token list filters, as they were kept in local state, so they now use `useFilterSearch` like the other lists.

### DIFF
--- a/src/pages/app/arch/list.tsx
+++ b/src/pages/app/arch/list.tsx
@@ -1,8 +1,7 @@
 import { ScrollArea } from "@/components/ui/scroll-area"
 
-import { cursorFromLink, useCursors } from "@/hooks/use-cursors"
+import { cursorFromLink, useCursorSearch } from "@/hooks/use-cursors"
 import { useArchTableColumns } from "@/hooks/use-arch-table-columns"
-import { useDataTable } from "@/hooks/use-data-table"
 import { Arch } from "@/types/arches"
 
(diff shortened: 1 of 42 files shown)


src/pages/app/arch/list.tsx
```diff
@@ -13,9 +12,8 @@ import PageHeader from "@/components/page-header"
 import PageFooter from "@/components/page-footer"
 
 export default function ArchesList() {
-  const table = useDataTable()
-  const { page, pageSize, setPage } = table
-  const { cursor, goToPage } = useCursors(page, setPage)
+  const table = useCursorSearch()
```
ezekg: Nit: still calling this var `table` throughout is confusing

ezekg (dismissed):

ezekg (approved):

PR #328, opened 2026-07-29
Add switch buttons to environments dialog

### LINKED ISSUE #325: Add change environment button to details view
Currently, the only place to switch to another environment is the dropdown:

Might get unruly if someone has a lot of environments.

Should add the ability to switch into an environment from its details view for better UX?

Cazden:
Closes #325.

### DIFF
--- a/src/components/environments/view/modal.tsx
+++ b/src/components/environments/view/modal.tsx
@@ -1,4 +1,4 @@
-import { useCallback } from "react"
+import { useCallback, useState } from "react"
 
 import { Button } from "@/components/ui/button"
 import { ScrollArea } from "@/components/ui/scroll-area"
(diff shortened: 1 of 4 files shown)


src/components/environments/view/modal.tsx
```diff
@@ -146,12 +193,30 @@ export default function EnvironmentsViewModal({
...
-            New Environment
-          </Button>
+          <div className="flex items-center gap-2">
+            {activeEnvironmentId != null && (
+              <Button
+                onClick={handleSwitchToGlobal}
+                disabled={switching}
+                className="min-w-36"
+              >
+                {switching ? <Loading.Dots /> : "Switch to Global"}
```
ezekg: Nit: I think we can do without this button since it's always at the top of the environment selector. Multiple primary buttons should usually be avoided, so if you do think we should keep it, make it secondary.

ezekg (dismissed):

src/components/environments/view/modal.tsx
Cazden: I removed it. I was already iffy about it but wanted to keep the theme of having the ability to switch environments from the dialog instead of just the dropdown. I think it's fine without though because like you said, global will always be at the top of the dropdown.

ezekg (approved):

Issue #120, opened 2026-03-26
Add ability to open resource details in new tab


ezekg:
Ctrl + clicking a resource from a list unintuitively opens the details page in the same tab. Likely going to need state persistence into local storage, e.g. current environment, etc. so that the new tab can navigate and hydrate correctly.

ezekg:
Docs: https://tanstack.com/router/v1/docs/guide/navigation

PR #329, opened 2026-07-29
Fix environment persistence

### LINKED ISSUE #326: Switching environments can cause Not Found errors
e.g. switching to an `Isolated` environment while on a resource's details page will subsequently cause a 404.

### LINKED ISSUE #327: Current environment should persist in local storage
Related to #236. Currently, environments aren't persisted to local storage, so refreshing the page will put the user back in `nil`.

Cazden:
Closes #327, closes #326, and also related to #120 as this adds environment persistence through session/local storage.


src/hooks/use-leave-details-route.ts
```diff
@@ -0,0 +1,18 @@
...
+  const router = useRouter()
+
+  return useCallback(async () => {
+    const { matches, location } = router.state
+    const routeId = matches.at(-1)?.routeId
+
+    if (routeId == null || !routeId.endsWith("/$id")) return
+
+    await router.navigate({
+      to: location.pathname.replace(/\/[^/]+$/, ""),
```
ezekg: Not sure the regex is needed here, we could probably just use `..`, no?

src/providers/environment-provider.tsx
```diff
@@ -41,17 +59,24 @@ function EeEnvironmentProvider({
...
+            token,
+          })
         }
 
         setId(environmentId)
         setCode(environmentCode)
+
+        // resource from the previous environment won't resolve in the new one,
+        // so nav off the details route to its list, if applicable
+        await leaveDetailsRoute()
```
ezekg: This really only needs to run on a 404 imo, because the resource _could_ still be visible, e.g. a shared env.

src/hooks/use-leave-details-route.ts
```diff
@@ -0,0 +1,18 @@
+import { useCallback } from "react"
+import { useRouter } from "@tanstack/react-router"
+
+export function useLeaveDetailsRoute(): () => Promise<void> {
```
ezekg: This name is a bit ambiguous, especially since it's called without condition from the environment provider, i.e. it isn't always called while inside of a details route. I would call this `useBackNavigateIfDetailsRoute` or something clearer, and maybe move it into `hooks/use-back-navigate.ts` too.

src/keygen/environment.ts
```diff
@@ -0,0 +1,84 @@
+import client from "@/keygen/client"
+import config from "@/keygen/config"
+
+const STORAGE_KEY = "environment"
```
ezekg: This needs a `keygen.` prefix like other storage keys.

src/providers/environment-provider.tsx
```diff
@@ -62,11 +87,72 @@ function EeEnvironmentProvider({
...
   )
 
+  // a restored environment may have been deleted or renamed while we were gone,
+  // which would scope every request to an environment that no longer resolves
+  useEffect(() => {
+    if (restored == null) return
+
+    let cancelled = false
+
+    void (async () => {
```
ezekg: This code is very messy and hard to reason about, especially with the IIFE. We should clean it up by extracting some of this logic out into separate pure functions and hooks.

src/keygen/environment.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  storage.setItem(STORAGE_KEY, JSON.stringify(stored))
+
+  client.setEnvironment(code)
+  client.setEnvironmentToken(token)
+}
+
+export function clearEnvironment(): void {
+  localStorage.removeItem(STORAGE_KEY)
+  sessionStorage.removeItem(STORAGE_KEY)
```
ezekg, line 65: Nit: I don't particularly like how the client is concerned with local/session storage. Storage seems like a separate concern to me i.e. I see the client more as something usable outside of React -- something we could eventually extract out into a Node package usable server-side, even, for example. But I see we already have this pattern in login/logout/session, so I'm leaving it here as a nit since it's largely unimportant right now.

src/keygen/login.ts
```diff
@@ -20,6 +21,10 @@ export function login(
   other.removeItem("tokenId")
   other.removeItem("token")
 
+  // new session starts in global environment, since any environment
+  // token or session from the previous session is no longer usable
+  clearEnvironment()
```
ezekg: Shouldn't this be called _before_ login, i.e. in authenticate?

src/providers/environment-provider.tsx
```diff
@@ -41,17 +59,24 @@ function EeEnvironmentProvider({
...
         if (environmentCode == null) {
-          keygen.client.setEnvironmentToken(null)
-          keygen.client.setEnvironment(null)
+          clearEnvironment()
         } else {
           const token = await getOrCreateEnvironmentToken(environmentId!)
 
-          keygen.client.setEnvironmentToken(token)
-          keygen.client.setEnvironment(environmentCode)
+          storeEnvironment({
```
ezekg: I feel like replacing these `keygen.client` calls kind of muddies the water on what's actually being done here. tbh, I think pushing the env storage calls into `setEnvironmentToken` and `setEnvironment` makes more sense, similar to what we're doing with login/logout/session, rather than exposing and using the "storage" functions here since we're not _just_ "storing" the environment, but _setting_ it on the client too.

ezekg (changes requested):
Have some concerns on code quality, as a lot of these changes have become hard to reason about. There's also a lack of separation of concerns, and too much implicit behavior, which I think makes the code harder to read as well.

I think the code needs a refactor and clean up pass to make behavior more explicit.

src/hooks/use-leave-details-route.ts
Cazden: I tried this but ran into some issues that warranted regex. `..` would resolve to `/app/dashboard` or even `/app`, showing no content outside of the sidebar.

src/keygen/environment.ts
Cazden: Opened #331 so we can track this later.

ezekg (approved):

PR #333, opened 2026-07-30
Fix default teammate permissions

### LINKED ISSUE #330: Permissions are unclear when inviting a teammate
We've had multiple support requests about teammates being unable to access Portal due to misconfigured permissions. I think this stems from the permissions defaulting to the bare-minimum when inviting a teammate, with little indication that additional permissions should be selected. We should adjust this UX by either selecting all by default, or adding more explicit messaging to choose additional permissions. I honestly vote for the former, with messaging that recommends customizing permissions where needed.

Image

Cazden:
Closes #330. Replaces default Portal required permission set with all permissions (based on current user's role; we wouldn't allow a non-admin to grant permissions that only an admin could) for the `Invite Teammate` form.


ezekg (changes requested):
Overall, looks good, and the messaging seems good. But I think we should default to an explicit set of permissions matching the current user vs "All." That way it'll be easier to simply click the "X" on permissions that aren't needed, vs deselecting "All" and then reselecting the desired permissions (and potentially missing a few required ones for Portal and causing an annoying back-and-forth UX).

ezekg (approved):

PR #342, opened 2026-08-20
Refactor permission select interface

### LINKED ISSUE #127: Redesign permission select to use dialog with descriptions
Right now, selecting permissions is a bit overwhelming. I'd like to redesign the permission select to open a dialog containing an organized set of permissions, with useful descriptions for each, as well as indicators for the default permission set. e.g.:

Image

### LINKED ISSUE #205: Default permissions are wrong in CE for some roles

Cazden:
Closes #127, closes #205.

Updates the permissions select UI to make it easier to manage permissions. Additionally, this also fixes the default permissions that were out of sync with API defaults.

For users that don't have the corresponding permission themselves and therefore can't grant it, they're disabled with a helpful tooltip:


src/components/users/form/fields.tsx
```diff
@@ -399,6 +401,87 @@ function LastNameField({
   )
 }
 
+function useAccountDefaultPermissions(): readonly string[] {
+  const { data: settings = [] } = useGetAccountSettings()
+
+  return useMemo(() => {
```
Cazden: All of these helper functions should be moved elsewhere, e.g. `/lib`. They shouldn't be in the form fields file.

src/components/permission-select.tsx
```diff
@@ -0,0 +1,577 @@
...
+            </CommandList>
+
+            <RequiredScrollHint listRef={listRef} />
+          </div>
+        </Command>
+
+        <div className="flex items-center gap-4 border-t border-accent p-3">
+          <span className="text-sm text-content-subdued">
+            {isNoneSelected
+              ? "No permissions"
```
Cazden: We should move this to the dialog description in its header, we don't usually show data in footers.

ezekg (changes requested):
Performance is much better, ty. The hover state of individual rows seems to be sticky, which I think should be fixed. Attached a video.

https://github.com/user-attachments/assets/f20fcf92-cd9d-4882-b1bf-5ad4c9dc7203

Cazden:
Fixed the sticky hover issue and improved UX for keyboard/tab navigation.

ezekg (approved):

PR #343, opened 2026-08-20
Add new account onboarding

### LINKED ISSUE #274: Add onboarding wizard for new accounts
New accounts are currently dropped onto the main metrics page, which is empty, and it's not super clear what they should do next (e.g. go to Licensing > Products > New). Instead, we should drop new accounts onto an onboarding page with a wizard that instructs them to invite teammates, create a product, a policy for it, and their first license using the policy, and then displays a `curl` snippet to validate the license (or validate it inline in the browser and display the JSON results, e.g. in a fake "terminal" kind of component).

Cazden:
Closes #274.

Adds an onboarding wizard and learning resource page that guides new users through inviting teammates, creating an environment, product, policy and license.

The quickstart wizard guides users through step-by-step, with options to skip inviting teammates or creating an environment:

Each form is a simplified version of its normal counterpart, with some helpful tips to make onboarding easier to follow:

Finally, users will finish onboarding with a license API validation that they can either perform in their own terminal or via our pseudo-terminal:


ezekg (changes requested):
As noted, the new permissions dialog seems to have some performance issues when checking/unchecking permissions, so I'd like to see that fixed before we merge as it could have other downstream effects. I also think that the quickstart's product/policy/license should be created within the environment, if one was created. Right now they're put into the global environment, which kind of defeats the purpose of having the user create an environment to separate things.

Outside of that, overall it looks good!

Aside: might be worth [sending onboarding-related events to Fathom](https://usefathom.com/docs/events/overview) (when Fathom is configured) so we can monitor how often users successfully onboard vs drop off at certain steps and adjust accordingly. Fine to do in a separate follow up (or stacked?) PR, ofc. Not a huge priority right now, more of a nice-to-have.

src/components/users/form/fields.tsx
```diff
@@ -893,3 +905,271 @@ function ConfirmPasswordField({
...
+  const form = useFormContext<Schemas.Users.InvitesValues>()
+  const role = useWatch({
+    control: form.control,
+    name: `invites.${index}.role`,
+  })
+
+  const { permissions: currentPermissions } = usePermissions()
+  const requiredOptions = useMemo(
+    () =>
+      PORTAL_REQUIRED_OPTIONS.filter((o) => currentPermissions.has(o.value)),
```
ezekg: When inviting an admin role, we also need to make sure `admin.read` is marked as required for Portal, otherwise the admin won't be able to read their own user: https://github.com/keygen-sh/keygen-api/blob/af351aed20101032348a3c146b6fdca7ece3b814/app/policies/user_policy.rb#L27

src/components/onboarding/form/environment.tsx
```diff
@@ -0,0 +1,168 @@
...
+  open,
+  onOpenChange,
+}: EnvironmentOnboardingFormProps) {
+  const form = useForm<Schemas.Environments.CreateValues>({
+    resolver: zodResolver(Schemas.Environments.CreateSchema),
+    mode: "onChange",
+    defaultValues: {
+      name: "",
+      code: "",
+      isolationStrategy: IsolationStrategy.Isolated,
```
ezekg: Let's create a shared environment here so our step detection doesn't get messed up if they create a product or policy out-of-band, e.g. right now if you create the environment, then leave onboarding and create a product, then return, the policy creation fails because the out-of-band product is in the global environment, but the policy is being created within the isolated onboarding environment. A shared environment would still allow this to succeed, since shared policies can be for a global product.

ezekg (changes requested):
A couple small changes, but other than that looks good! One more thing: let's either hide the learn page and redirect it to metrics when inside of an environment, or make sure we're querying the global environment during onboarding. I feel like the former is easier and makes more sense. Right now if you onboard and then switch environments, it restarts.

Cazden:
Onboarding page is now exclusive to the global environment and redirects/hides the `/learn` route if not in global. Removed the option to select `isolated`, so the user is automatically funneled into creating a `shared` environment, and all resources created *within* the onboarding wizard are now created in that `shared` environment, if one was created.

src/components/onboarding/dialog/validation.tsx
```diff
@@ -0,0 +1,350 @@
...
+        }
+      }
+    },
+    [validateKey],
+  )
+
+  const handleCommand = useCallback(
+    (input: string): string | Promise<string> => {
+      if (input.includes("halo")) {
+        return "keygen: I think Halo is a pretty cool guy. Eh kills aleins and doesnt afraid of anything."
```
ezekg, line 160: Image

ezekg (approved):
Looks good! Excited to see how it improves onboarding!

PR #347, opened 2026-08-27
Add Fathom event tracking to onboarding wizard

Cazden:
This will add event analytics to our Fathom dashboard for opening, skipping and 
completing any step in our onboarding wizard flow. If sending events for opening the steps seems like it'd create too much noise, I can remove that, but also easy to do in a follow-up if you want to see how it does.


ezekg (dismissed):

Cazden:
@ezekg Will need reapproval on this once the other branches in the stack are approved.

ezekg (approved):

PR #352, opened 2026-09-02
Add CSV export to license expiration heatmap

### LINKED ISSUE #348: Add CSV export functionality to expiring license heatmap
Customers have mentioned it'd be useful to export expiring licenses into a CSV file for their sales teams.

Exporting the next year of expirations would be fine, I think.

Cazden:
Closes #348. Adds an actions dropdown button to the license expiration heatmap with (currently) a single option to export expiring licenses as a CSV.

### DIFF
--- a/src/components/analytics/license-expiration-heatmap.tsx
+++ b/src/components/analytics/license-expiration-heatmap.tsx
@@ -20,12 +20,12 @@ import { Skeleton } from "@/components/ui/skeleton"
 
 import * as keygen from "@/keygen"
 
-import { cn } from "@/lib/utils"
 import {
   toDisplayRow,
   getTemperatureColor,
   buildExpirationHeatmap,
 } from "@/lib/analytics"
+import { cn } from "@/lib/utils"
 import { truncator } from "@/lib/truncate"
 
 import { License } from "@/types/licenses"
@@ -33,6 +33,7 @@ import { ExpirationHeatmapEntry } from "@/types/analytics"
 
 import {
   useExpirationsHeatmap,
+  useExportExpiringLicenses,
   useLicensesExpiringOn,
 } from "@/queries/analytics"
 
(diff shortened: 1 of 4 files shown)


src/queries/analytics.ts
```diff
@@ -222,6 +228,58 @@ export function useLeaderboard(
...
+            environment: code,
+          })
+
+          if (response.errors) {
+            throw new APIError(response.errors[0])
+          }
+
+          licenses.push(...(response.data ?? []))
+          cursor = cursorFromLink(response.links?.next)
+        } while (cursor)
```
ezekg: This could get out of hand for large accounts with thousands of licenses (some even have hundreds of thousands). I want to add a max pagination limit of 10 pages before we do this, e.g. if the heatmap returns a summed count of over 1k licenses, export should fail with an error and instruct the user to select a smaller export window, e.g. 90 days. This should help prevent performance and rate limiting issues.

ezekg (changes requested):

src/queries/analytics.ts
```diff
@@ -229,55 +229,82 @@ export function useLeaderboard(
...
         }
 
-        downloadCsv(licensesToCsv(licenses), filename)
+        licenses.push(...(response.data ?? []))
+        cursor = cursorFromLink(response.links?.next)
+        pages += 1
+      } while (cursor && pages < MAX_EXPORT_PAGES)
+
+      if (cursor) {
+        throw new ExportLimitError()
```
ezekg: I say we don't raise an error here, as the export has already partially-completed, so we should at least provide what we have, otherwise we just wasted a bunch of requests. Maybe download, but with a message indicating that the export did not fully complete because it exceeded the total number of licenses allowed in an export.

src/components/analytics/license-expiration-heatmap.tsx
```diff
@@ -109,7 +110,8 @@ export default function LicenseExpirationHeatmap({
     const exportEnd = format(addDays(new Date(), rangeDays - 1), "yyyy-MM-dd")
 
     exportLicenses.mutate({
-      before: `${exportEnd}T23:59:59.999Z`,
+      before: endOfDayUtc(exportEnd),
+      count: cells.reduce((sum, cell) => sum + cell.count, 0),
```
ezekg: I feel like the `count` here and the subsequent error guard are a bit misplaced. The error should probably be thrown from the `handleExport()` handler rather than from `exportLicenses.mutate()`. It seems a little weird to pass in a count solely to check whether we should raise an error, especially when the export hook itself doesn't really care about the count. It's more that the caller cares that we don't try to export too many licenses.

ezekg (approved):
Sick! LGTM!

PR #362, opened 2026-09-10
Fix relationship failures showing as failed creates

### LINKED ISSUE #358: Failure to create license still creates it under certain conditions
For example, when attaching an entitlement that's already attached via policy, the license is still created. Should either bail creating the license if entitlement attachment fails, or still close the form/ignore the attachment if we receive that code specifically since it's already attached via policy as the user requested. Might even be a good idea to prune it from the list and add it as a locked selection so the user knows it's already there.

Right now,  we're 1) incorrectly notifying the user that license creation failed (it did not):

And 2) not closing the form is deceiving because the user likely wouldn't realize the license was created until after it's closed, resulting in situations like what happened to me, where I accidentally created duplicate licenses, as I assumed the first one failed:

Cazden:
Closes #358.


src/components/licenses/form/create.tsx
```diff
@@ -77,46 +77,54 @@ export default function CreateLicenseForm({
...
-      }
-
-      if (createdEntitlementIds.length > 0)
-        await attachEntitlements.mutateAsync({
-          licenseId: license.id,
-          entitlementIds: createdEntitlementIds,
-        })
-
+      const ownerId = values.ownerId
+      const groupId = values.groupId
```
ezekg: This could probably be simplified using destructuring:

```ts
const { ownerId, groupId } = values
```

src/components/machines/form/create.tsx
```diff
@@ -56,21 +56,25 @@ export default function CreateMachineForm({
...
-      if (values.groupId) {
-        await changeGroup.mutateAsync({
-          machineId: machine.id,
-          groupId: values.groupId,
-        })
-      }
-
-      toast({ message: "Machine activated", variant: "success" })
+      const ownerId = values.ownerId
+      const groupId = values.groupId
```
ezekg: Same comment

Cazden: Oops.

ezekg (approved):

PR #363, opened 2026-09-10
Add inherited entitlements to license creation

Cazden:
Adds better UX during license creation by showing inherited entitlements from the selected policy in the license creation form.


src/components/licenses/form/create.tsx
```diff
@@ -74,6 +74,8 @@ export default function CreateLicenseForm({
     () => policies.find((p) => p.id === selectedPolicyId) ?? null,
     [policies, selectedPolicyId],
   )
+  const { data: policyEntitlements = [] } =
+    useListPolicyEntitlements(selectedPolicyId)
```
ezekg: This should probably have a high limit of `100` instead of the default `10`, since some policies have tens of entitlements. Fine to cap the query at 100 and not paginate to reduce cost of what amounts to a UX fix.

src/queries/policies.ts
```diff
@@ -163,7 +163,7 @@ export function useListPolicyEntitlements(policyId: string) {
     queryKey: ["policies", policyId, "entitlements", { environment: code }],
     queryFn: () =>
       keygen.policies
-        .listEntitlements({ policyId })
+        .listEntitlements({ policyId, limit: 100 })
```
ezekg: Is this going to effect the policy details page? I don't remember how that part is set up, e.g. if it uses pagination or what. Should probably be a parameter if so because showing 100 entitlements is a bit much.

Image

ezekg (approved):

PR #372, opened 2026-09-17
Add support for static links

Cazden:
Adds support for static/account-agnostic links to pages within Portal, e.g. `/goto/licenses` will redirect unauthenticated users to login with `?redirect=/goto/licenses`, and upon successful authentication, will use that param to send them back through `/goto/licenses`, which will take them to `/<account>/app/licenses`.


src/components/auth/form/login.tsx
```diff
@@ -64,6 +65,11 @@ export default function LoginForm() {
 
     session.setUser(userId)
 
+    if (redirect) {
+      void navigate({ href: redirect })
```
ezekg: This and other `navigate` cases should use `to:`, not `href:`, otherwise we could introduce an open redirect vulnerability, e.g. `?redirect=https://evil.example`. As far as I can tell, the `href:` param should only be used for safe external navigation, and never with user-controlled input. See [Tanstack docs](https://tanstack.com/router/latest/docs/api/router/RedirectType) for more.

ezekg (approved):

